PR #2, opened 2019-07-23
Bugfix qa inference mode, change layer dims handling

Timoeller:


tholor:
Looks good! I have just simplified the input format for the QA example a bit. We don't need the language parameter there.

PR #5, opened 2019-07-23
Add multiGPU support for experiments and examples

Timoeller:
As easy as adding:
self.prediction_heads = nn.ModuleList([ph.to(device) for ph in prediction_heads])

Eval mode should also work now for prediction heads.

### DIFF
--- a/examples/doc_classification.py
+++ b/examples/doc_classification.py
@@ -24,7 +24,7 @@ ml_logger.init_experiment(experiment_name="Public_FARM", run_name="Run_doc_class
 ########## Settings
 ##########################
 set_all_seeds(seed=42)
-device, n_gpu = initialize_device_settings(use_cuda=True, local_rank=-1, fp16=False)
+device, n_gpu = initialize_device_settings(use_cuda=True)
 n_epochs = 1
 batch_size = 32
 evaluate_every = 30
--- a/examples/doc_classification.py
+++ b/examples/doc_classification.py
@@ -1,18 +1,16 @@
 # fmt: off
 import logging
 
-import torch
-
 from farm.data_handler.data_silo import DataSilo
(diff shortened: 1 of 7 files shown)


examples/doc_classification.py
```diff
@@ -26,7 +24,7 @@
 ########## Settings
 ##########################
 set_all_seeds(seed=42)
-device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+device, n_gpu = initialize_device_settings(use_cuda=True, local_rank=-1, fp16=False)
```
tholor: I see why you are adding this here for multi-gpu support, but I think we should keep the examples rather simple. Can you make local_rank=-1 and fp16=False default args of  initialize_device_settings? Then we don't need to list them in the examples.

farm/modeling/adaptive_model.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         """
         super(AdaptiveModel, self).__init__()
         self.language_model = language_model.to(device)
-        self.prediction_heads = [ph.to(device) for ph in prediction_heads]
+        self.prediction_heads = nn.ModuleList([ph.to(device) for ph in prediction_heads])
```
tholor: Have you tested training, saving and loading for the examples after introducing this?

examples/doc_classification.py
Timoeller: good idea, done.

farm/modeling/adaptive_model.py
Timoeller: I have tested training on all tasks.
I haven't tested saving, but this shouldn't impact saving at all.
I loaded a QA model that was saved with the old code and it worked. Haven't tested the other models.

PR #6, opened 2019-07-24
tutorial in working state

brandenchan:
Could someone please have a read of this notebook, run it and let me know if any bugs come up? I managed to make it work on cpu and cuda.


Timoeller:
I tested it on my machine, works like a charm.
Except the GPU memory isnt released when stopping a cell. But that will be more of a jupyter issue? Restarting the kernel clears the memory.

PR #13, opened 2019-07-26
Fix docs

brandenchan:
A first pass at cleaning up the docs


brandenchan:
A first pass at cleaning up the docs

Issue #16, opened 2019-07-29
Initialize Inferencer from building blocks


tholor:
Currently the inferencer is always constructed by loading a saved model from disk:
https://github.com/deepset-ai/FARM/blob/a537b25c9bf604c49acc4a7414ab52c56aac4ab9/farm/infer.py#L30-L58

Let's change it to have a classmethod "load" for doing this and a constructor expecting an adaptive model + processor.

PR #18, opened 2019-07-29
Add load_from_dir to Inferencer

busyxin:
This PR addresses #16 
Some observations though:

1. When using the regular constructor (not from dir), what name property should the Inferencer class get assigned? `Infer.py L.59`

2. Any suggestions to not using `initialize_device_settings` in both constructor and `load_from_dir`?


farm/infer.py
```diff
@@ -54,9 +56,28 @@ def __init__(self, load_dir, batch_size=4, gpu=False):
             self.label_map = self.processor.label_maps[0]
         elif len(self.model.prediction_heads) == 0:
             self.prediction_type = "embedder"
-        self.name = os.path.basename(load_dir)
+        # self.name = os.path.basename(load_dir)
         set_all_seeds(42, n_gpu)
 
+    @classmethod
+    def load_from_dir(cls, load_dir, gpu=False):
```
tholor: Please name it "load" to be more consistent with the rest of the package

farm/infer.py
```diff
@@ -54,9 +56,28 @@ def __init__(self, load_dir, batch_size=4, gpu=False):
...
+        :return: An instance of the Inferencer.
+        """
+
+        device, n_gpu = initialize_device_settings(
+            use_cuda=gpu, local_rank=-1, fp16=False
+        )
+
+        model = AdaptiveModel.load(load_dir, device)
+        processor = Processor.load_from_dir(load_dir)
+        return cls(model, processor)
```
tholor: Pass the parameter "gpu" to the constructor

farm/infer.py
```diff
@@ -54,9 +56,28 @@ def __init__(self, load_dir, batch_size=4, gpu=False):
             self.label_map = self.processor.label_maps[0]
         elif len(self.model.prediction_heads) == 0:
             self.prediction_type = "embedder"
-        self.name = os.path.basename(load_dir)
+        # self.name = os.path.basename(load_dir)
```
tholor: Suggestion for the name: 
- We make it an optional argument in the constructor
- In case of loading from dir, we still pass the directory name to it
- In case of initializing from building blocks, users can supply a name if they want
- If no name is supplied: the name is set to `self.name = f"anonymous-{self.prediction_type}"`

examples/embeddings_extraction.py
```diff
@@ -0,0 +1,48 @@
...
+batch_size = 32
+lang_model = "bert-base-german-cased"
+
+# 1.Create a tokenizer
+tokenizer = BertTokenizer.from_pretrained(
+    pretrained_model_name_or_path=lang_model, do_lower_case=False
+)
+
+# 2. Create a DataProcessor that handles all the conversion from raw text into a pytorch Dataset
+processor = GNADProcessor(
```
tholor: This is fine for now, but we will need to switch this processor here to a more generic one. Something like a "BasicProcessor". Since we already had something similar in mind to simplify most of the text classification processors, I wouldn't do it right now here, but in a separate MR later.

farm/infer.py
busyxin: `load_from_dir` can be found in `processor.py` and it is also mentioned in `prediction_head.py` (but seems like an outdated comment).

PR #20, opened 2019-07-30
Optional evals

brandenchan:
Evaluators are now optional. The processor will not fetch a dev set if dev_filename=None and dev_split = 0.0. The processor will not fetch a test set if test_filename=None.

### DIFF
--- a/farm/train.py
+++ b/farm/train.py
@@ -96,14 +96,14 @@ class Trainer:
         self.learning_rate = self.optimizer.get_lr()
         self.warmup_linear = warmup_linear
         self.global_step = 0
-        self.data_loader_train = data_silo._get_data_loader("train")
+        self.data_loader_train = data_silo.get_data_loader("train")
         self.device = device
         self.log_params()
 
         # evaluator on dev set
-        if evaluator_dev is None:
+        if evaluator_dev is None and self.data_silo.get_data_loader("dev"):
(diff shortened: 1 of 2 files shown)


farm/train.py
```diff
@@ -96,24 +96,24 @@ def __init__(
...
         self.warmup_linear = warmup_linear
         self.global_step = 0
-        self.data_loader_train = data_silo._get_data_loader("train")
+        self.data_loader_train = data_silo.get_data_loader("train")
         self.device = device
         self.log_params()
 
         # evaluator on dev set
-        if evaluator_dev is None:
+        if evaluator_dev is None and self.data_silo.get_data_loader("dev"):
```
Timoeller: I was wondering about this line. Why do we need to have evaluator_dev and _test in the init method of Trainer, when we have these guys inside our data_silo (and a Trainer always needs a data_silo)?
I checked the code occurrences and we never pass an evaluator object anyways. So lets get rid of it?

brandenchan: The requirements I had in mind for the trainer are 1) that evaluators can be passed to its constructor, 2) that if no evaluator is passed, it can be automatically initialized in the trainer, 3) we can run the trainer without any evaluator. The double condition in this line allows for us to achieve 3) since the data_silo can be configured to not produce a dev or test loader

PR #24, opened 2019-07-30
ner - f1 - preprocessing predictions before f1_score calculation: rem…

waldemarhahn:
Removing X and [PAD] - tokens, removing predictions beginning with I-token -> boosts performance for NER

### DIFF
--- a/farm/eval.py
+++ b/farm/eval.py
@@ -7,12 +7,21 @@ from seqeval.metrics import classification_report as token_classification_report
 from sklearn.metrics import classification_report
 from torch.utils.data import DataLoader
 
-from farm.metrics import compute_metrics
+from farm.metrics import compute_metrics, _correct_bio_encodings
 from farm.utils import to_numpy
 from farm.utils import MLFlowLogger as MlLogger
 from farm.modeling.adaptive_model import AdaptiveModel
 
 logger = logging.getLogger(__name__)
+
+
+def _token_classification_report(labels, preds, digits):
+    preds = [
+        [token if token not in {"X", "[PAD]"} else "O" for token in sent]
+        for sent in preds
+    ]
+    preds = _correct_bio_encodings(preds)
+    return token_classification_report(labels, preds, digits)
 
 
 class Evaluator:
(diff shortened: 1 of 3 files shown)


farm/eval.py
```diff
@@ -7,14 +7,23 @@
...
 from farm.utils import to_numpy
 from farm.utils import MLFlowLogger as MlLogger
 from farm.modeling.adaptive_model import AdaptiveModel
 
 logger = logging.getLogger(__name__)
 
 
+def _token_classification_report(labels, preds, digits):
+    preds = [
+        [token if token not in {"X", "[PAD]"} else "O" for token in sent]
```
Timoeller: Are you sure this is still needed? Even when removing these tokens from our label list?

farm/eval.py
```diff
@@ -7,14 +7,23 @@
...
+        for sent in preds
+    ]
+    preds = _correct_bio_encodings(preds)
```
Timoeller: I would either move this function in token_classification_report() or rename _token_classification_report() to corrected_token_classification_report() to not confuse people.

farm/metrics.py
```diff
@@ -59,6 +61,40 @@ def compute_metrics(metric, preds, labels):
         raise KeyError(metric)
 
 
+def _correct_bio_encodings(predictions):
```
Timoeller: Could you 
- make this function non private (I think people would like to use it regularly) 
- comment on the reasoning why we are doing this and how it is done
Thanks

farm/eval.py
```diff
@@ -7,14 +7,23 @@
...
-from farm.metrics import compute_metrics
+from farm.metrics import compute_metrics, _correct_bio_encodings
 from farm.utils import to_numpy
 from farm.utils import MLFlowLogger as MlLogger
 from farm.modeling.adaptive_model import AdaptiveModel
 
 logger = logging.getLogger(__name__)
 
 
+def _token_classification_report(labels, preds, digits):
```
tholor: Please test if this report also still works with LM finetuning. There we use it to report performance on the masked language model task. If it doesn't work we might need two separate report types here.

Timoeller:
Hey @waldemarhahn 
Do you need any support on the changes we suggested or have comments about them? If you do not have time, I will take over from your codebase by the end of this week to finish this PR.
Thanks again for the improvements you implemented.

waldemarhahn:
Hey @Timoeller
Sorry for not replying for that long, unfortunately I don't have time for fixing this right now... I would be happy, if you could "finish" it and I think a function for transforming IOB to BIO format (for CoNNL dataset) should be useful. Otherwise this tweak won't help to improve the stats for that dataset.

Timoeller:
I am closing this PR because of inactivity.

Issue #30, opened 2019-08-01
Accuracy metric in LM finetuning always zero


tholor:
The accuracy metric reported by the Evaluator for masked LM task during LM finetuning is wrong. This is due to the fact that we have there a nested list of predictions with different lengths (different number of masked tokens per sample). This leads to wrongly converted numpy arrays and a wrong metric. I am working on a fix.

tholor:
Fixed by https://github.com/deepset-ai/FARM/pull/31

PR #31, opened 2019-08-01
fix accuracy metric for nested lists of diff lengths

tholor:
This adresses issue https://github.com/deepset-ai/FARM/issues/30

The suggested fix is to pad the lists to equal length before converting to a numpy array and then ignoring the PAD elements in the calculation of the metric.

### DIFF
--- a/farm/metrics.py
+++ b/farm/metrics.py
@@ -8,10 +8,24 @@ from sklearn.metrics import matthews_corrcoef, f1_score
 def simple_accuracy(preds, labels):
     # TODO: THIS HACKY TRY CATCH IS FOR GNAD
     try:
+        # nested lists with different lengths
+        if type(preds[0]) == list:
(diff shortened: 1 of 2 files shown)


farm/metrics.py
```diff
@@ -8,10 +8,24 @@
 def simple_accuracy(preds, labels):
     # TODO: THIS HACKY TRY CATCH IS FOR GNAD
     try:
+        # nested lists with different lengths
```
Timoeller: I'd prefer more simple ways of doing this. I think flattening both preds or labels as 1d and then doing accuracy should do the same without padding.

tholor: Ok, good idea :+1: 
I will change it to 1d

tholor:
Changed it as suggested to a flat list conversion. That's way cleaner now.  Please review and merge @Timoeller

Issue #39, opened 2019-08-04
output_dir parameter in run by config is being ignored


tripl3a:
If you set the parameter like this:
```
    "output_dir": {"value": "/path/to/output", "default": "", "desc": "Output directory where model predictions and checkpoints will be saved."},
```
You will get an error if the directory already exists and the run will stop.

If the director does not exist the run will be successfully executed but the directory will be empty afterwards and the log will contain such entries:
```
08/03/2019 15:11:20 - WARNING - farm.file_utils -   Path saved_models/bert-german-GermEval18Coarse already exists. You might be overwriting files.
```

tripl3a:
Also `dev_size`is being ignored. Trying the three following configurations always defaulted to `dev_size=0.1`.

```
    "dev_size":         {"value": [0.2,0.3], "default": 0.1, "desc": "Split a dev set from the training set using dev_size as proportion."}
    "dev_size":         {"value": 0.2, "default": 0.1, "desc": "Split a dev set from the training set using dev_size as proportion."}
    "dev_size":         {"value": null, "default": 0.2, "desc": "Split a dev set from the training set using dev_size as proportion."}
```

tanaysoni:
Hi @tripl3a. Thank you for reporting the issue. It is resolved by #43.

PR #43, opened 2019-08-05
Fix `output_dir` param in configs (#39)

tanaysoni:
This PR implements use of the `output_dir` param and updates it's default value in configs.

### DIFF
--- a/farm/file_utils.py
+++ b/farm/file_utils.py
@@ -388,10 +388,3 @@ def unnestConfig(config, flattened=False):
             unnestedConfig.append(tempconfig)
 
     return unnestedConfig
-
-
-def create_folder(dir):
-    if os.path.exists(dir):
-        logger.warning(f"Path {dir} already exists. You might be overwriting files.")
-    else:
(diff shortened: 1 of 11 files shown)


farm/file_utils.py
```diff
@@ -388,10 +388,3 @@ def unnestConfig(config, flattened=False):
             unnestedConfig.append(tempconfig)
 
     return unnestedConfig
-
-
-def create_folder(dir):
-    if os.path.exists(dir):
-        logger.warning(f"Path {dir} already exists. You might be overwriting files.")
```
Timoeller: I think it makes sense to at least sent a notification if you potentially overwrite files. 

The problem currently is that this code is called twice when saving a model (processor.save and adaptive_model.save). So this bit is always executed, even when saving to a new folder. We should check weather the folder contains files and then print the warning.

btw: os.makedirs(exist_ok) seems to be removed from Python 3.4.1 onwards. https://bugs.python.org/issue21082

PR #62, opened 2019-08-15
Merge Processors, adjust config to be nested

Timoeller:
This PR merges some processors that were redundant
- Processors are more configurable now
- germeval18 Fine + Coarse and GNAD became a TextClassificationProcessor
- CoNLL03 and GermEval14 became NERProcessor
- The config is not flattened any longer, it is a dict of dicts now.

- examples are working
- experiments are working for germeval18Coarse and conll03
- all test pass (but LM finetuning has no loading and inference inside test)

Changes for configs Germeval28Fine, GNAD and GermEval14 are missing


farm/file_utils.py
```diff
@@ -284,13 +284,6 @@ def get_file_extension(path, dot=True, lower=True):
...
 
 
-# TODO: still needed?
-# Allow retrieval of config args with dot notation
-class Struct:
-    def __init__(self, **entries):
-        self.__dict__.update(entries)
-
-
 def read_config(path, flattend=False):
```
tholor: Should we move the config related functions to "experiment.py"? or an own config.py?

farm/data_handler/processor.py
```diff
@@ -656,6 +465,8 @@ def __init__(
         # Custom attributes
         self.delimiter = ""
 
+        self.additional_params = kwargs
```
tholor: For what is additional_params needed here?

farm/data_handler/processor.py
```diff
@@ -129,15 +130,12 @@ def load(cls, processor_name, data_dir, tokenizer, max_seq_len, dev_split=0):
...
+        :param kwargs: placeholder for passing generic parameters
+        :type kwargs: object
         :return: An instance of the specified processor.
         """
         return cls.subclasses[processor_name](
-            data_dir=data_dir,
-            tokenizer=tokenizer,
-            max_seq_len=max_seq_len,
-            dev_split=dev_split,
+            data_dir=data_dir, tokenizer=tokenizer, max_seq_len=max_seq_len, **kwargs
```
tholor: Why are we moving dev_split into **kwargs? Isn't that an arg for *every* processor?

farm/data_handler/processor.py
```diff
@@ -279,50 +289,44 @@ def _log_params(self):
...
         data_dir,
-        train_filename="train.csv",
-        dev_filename=None,
-        test_filename="test.csv",
-        dev_split=0.1,
+        train_filename="train.tsv",
+        dev_filename="dev.tsv",
+        test_filename="test.tsv",
+        dev_split=None,
+        label_list=[],
```
tholor: make this a mandatory arg?

farm/data_handler/processor.py
```diff
@@ -730,6 +542,8 @@ def __init__(
         :type doc_stride: int
         :param max_query_length: Maximum length of the question (in number of subword tokens)
         :type max_query_length: int
+        :param kwargs: placeholder for passing generic parameters
+        :type kwargs: object
```
tholor: - we should move label_list into the constructor as an arg with default values
 - we don't need to set dev_filename etc here (it's passed to the parent class and gets set there)

Issue #66, opened 2019-08-19
Unwanted type conversion in Class TextClassificationProcessor


AhmedIdr:
Im not sure if this a bug or not, but im getting an error when using the doc_classification example for custom dataset labeled True and False. I managed to solve it by just adding label before true and false, but just wanted to report it anyway.

Timoeller:
Thanks @AhmedIdr for using Farm. 
Could you provide more information on what kind of custom dataset you are doing doc_classification? And also what kind of error is thrown. Because I suspect this to be a hidden bug.

Potential sources for errors are:
1. Your custom dataset has a different format, especially a different separator than the default "\t" tab one?
2. Farm doesn't handle booleans correctly. We use pandas read_csv method, that will convert your "True" and "False" labels to booleans, but not if you manually change them to "label_True" and "label_False". Having booleans as labels might produce errors further down the line.

Please paste the code you are using and the error message, so we can potentially fix that issue.
Cheers, Timo

AhmedIdr:
Hey @Timoeller, Thank you for your answer. I am using the GermEval 2017 Dataset (https://sites.google.com/view/germeval2017-absa/home), the doc_classification example worked when using the sentiment label but doesn't work with the relevance (True, False), I checked the type of the labels and they are numpy.bool_ , so your second guess seems to be right. 
This is the only part I changed in the example: 
<img width="561" alt="doc_class" src="https://user-images.githubusercontent.com/31652778/63277552-716af600-c29d-11e9-9ba5-90c4bb6dec75.png">
and this is the error message:
<img width="907" alt="Farm_error" src="https://user-images.githubusercontent.com/31652778/63277634-952e3c00-c29d-11e9-9c77-286214585189.png">
I also noticed that it doesn't show the label_ids when reading the data and showing the sample from each file.

Timoeller:
Nice! Found a bug : )
Your label_list has strings, but the pandas function automatically converts the label column into appropriate formats. This is clearly unwanted.

If you want, you could fix that by always casting the input to a string and create a PR. It's happening inside this function: https://github.com/deepset-ai/FARM/blob/master/farm/data_handler/utils.py#L31

If you dont have time I will do it sometime later this week myself.

@busyxin Please also note this change. I guess your regression labels need conversion to ints or floats afterwards. I still think having one return type (=String) would be better, and let individual processors deal with the right format.

AhmedIdr:
Thank you for your Feedback, I might take a look at it later and fix it.

PR #67, opened 2019-08-19
Label Type casting in TextClassificationProcessor

AhmedIdr:
Casting all labels to strings to suppress the type conversion from pandas when reading files. 
This fixes issue #66

### DIFF
--- a/farm/data_handler/utils.py
+++ b/farm/data_handler/utils.py
@@ -35,8 +35,8 @@ def read_tsv(filename, quotechar='"', delimiter="\t", skiprows=None, columns=Non
         quotechar=quotechar,
         names=columns,
         skiprows=skiprows,
+        dtype=str, 
     )
-    df['label'] = df['label'].astype(str)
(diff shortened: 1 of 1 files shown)


farm/data_handler/utils.py
```diff
@@ -36,6 +36,7 @@ def read_tsv(filename, quotechar='"', delimiter="\t", skiprows=None, columns=Non
         names=columns,
         skiprows=skiprows,
     )
+    df['label'] = df['label'].astype(str)
```
Timoeller: Hey that is one way of doing it, that will solve the problem. I think it is slightly better to give a dtype=str argument to the pandas.read_csv directly, the automatic casting seems like wasted FLOPs : )
And we actually also want the input text fields to always be strings, so the dtype=str should be applied to all columns (and not just the label column).

AhmedIdr: I just changed it as you requested. : )

PR #75, opened 2019-08-30
WIP: Multiple text class heads

brandenchan:
Processors now have a tasks attribute that define the link from output tensor to prediction head


tholor:
@tanaysoni since here are many big changes happening, can you please check if the inferencer still works as before?

Issue #70, opened 2019-08-28
'NERProcessor' has no attribute 'tokenizer'


JulianGerhard21:
Python: 3.6.8
OS: Windows 10
FARM: 0.2.0

Hi guys,

if I want to run FARM to do some NER finetuning, I tried both approaches: 

`experiments = load_experiments("conll_de_config.json")`

and the stick-your-blocks-together one. Both of them ended with the following error:

```
08/28/2019 07:47:35 - INFO - farm.utils -   device: cuda n_gpu: 1, distributed training: False, 16-bits training: False
08/28/2019 07:47:35 - INFO - pytorch_transformers.tokenization_utils -   loading file https://int-deepset-models-bert.s3.eu-central-1.amazonaws.com/pytorch/bert-base-german-cased-vocab.txt from cache at C:\Users\JulianGerhard\.cache\torch\pytorch_transformers\da299cdd121a3d71e1626f2908dda0d02658f42e925a3d6abd8273ec08cf41a6.2a48e6c60dcdb582effb718237ce5894652e3b4abb94f0a4d9a857b70333308d
08/28/2019 07:47:35 - INFO - farm.data_handler.data_silo -   
Loading data into the data silo ... 
              ______
               |o  |   !
   __          |:`_|---'-.
  |__|______.-/ _ \-----.|       
 (o)(o)------'\ _ /     ( )      
 
08/28/2019 07:47:35 - INFO - farm.data_handler.data_silo -   Loading train set from: data/conll03-de\train.txt 
08/28/2019 07:47:35 - INFO - farm.data_handler.utils -    Couldn't find data/conll03-de\train.txt locally. Trying to download ...
08/28/2019 07:47:35 - INFO - farm.data_handler.utils -   downloading and extracting file C:\Users\JulianGerhard\PycharmProjects\word_embeddings\ner_finetuning\data\conll03-de to dir 
08/28/2019 07:47:39 - INFO - farm.data_handler.processor -   Got ya 8 parallel workers to fill the baskets with samples (chunksize = 1000)...
  0%|          | 0/24000 [00:00<?, ?it/s]multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "C:\Users\JulianGerhard\AppData\Local\Programs\Python\Python36\lib\multiprocessing\pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
[...]

PR #76, opened 2019-08-30
Add flag to disable multiprocessing (#70)

tanaysoni:
Currently, the Processor Class use `fork` to create new processes that isn't compatible with Windows OS. The flag allows disabling multiprocessing when running on Windows or for debugging purposes.


farm/data_handler/processor.py
```diff
@@ -71,6 +71,7 @@ def __init__(
         multiprocessing_chunk_size=1_000,
         max_processes=128,
         share_all_baskets_for_multiprocessing=False,
+        use_mutliprocessing=True
```
tholor: typo?

PR #81, opened 2019-09-05
Update comments to include tasks

brandenchan:


farm/data_handler/input_features.py
```diff
@@ -131,7 +131,6 @@ def samples_to_features_ner(
     max_seq_len,
     tokenizer,
     cls_token="[CLS]",
```
tholor: As I see it we currently don't use any of those special token args. We should use cls_token and sep_token instead of the static strings in the code  ...

PR #83, opened 2019-09-09
Hotfix to load multiple inferencers

tanaysoni:
During initialisation of `Processor` class, some class attributes gets updated. It leads to new `Processor` objects overwriting the properties of any existing `Processor` instances.

As an intermediary fix to support multiple inferencers, the REST API reloads an `Inferencer` each time it gets selected for inference.


tanaysoni:
This issue is resolved by #88.

PR #89, opened 2019-09-13
Multilabel Classification

tholor:
Allowing mutlilabel classification by re-using TextClassificationProcessor and adding a MultiLabelTextClassificationHead


tholor:
@Timoeller please review and test. Thanks!

PR #107, opened 2019-10-04
Add method to run inference from a file

tanaysoni:
The current `Inferencer` only supports running inference from dicts. 

This PR adds a new method `inference_from_file()` that takes a file as an input and uses multiprocessing to do speed-up inference preprocessing.

Additionally, the `_file_to_dicts()` in the `Processor` is made public.

### DIFF
--- a/farm/infer.py
+++ b/farm/infer.py
@@ -1,8 +1,14 @@
 import os
 import torch
 import logging
+import multiprocessing as mp
+from contextlib import ExitStack
+from functools import partial
 
+
+from torch.utils.data import ConcatDataset
 from torch.utils.data.sampler import SequentialSampler
+from tqdm import tqdm
 
 from farm.data_handler.dataloader import NamedDataLoader
 from farm.modeling.adaptive_model import AdaptiveModel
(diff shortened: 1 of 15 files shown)


farm/infer.py
```diff
@@ -110,27 +119,48 @@ def load(cls, load_dir, batch_size=4, gpu=False, embedder_only=False, return_cla
...
+
+            concat_datasets = ConcatDataset(datasets)
+
+        preds_all = self._run_inference(concat_datasets, tensor_names, samples)
+        return preds_all
+
+    @classmethod
+    def _multiproc_dict_to_samples(cls, dicts, processor):
+        dicts_list = [dicts]
+        dataset, tensor_names = processor.dataset_from_dicts(dicts_list, from_inference=True)
```
tholor: This means we call dict_to_samples twice. Once for "dataset_from_dicts" and once directly. If we want to improve speed, we probably want to change this to only be called once

farm/infer.py
```diff
@@ -110,27 +119,48 @@ def load(cls, load_dir, batch_size=4, gpu=False, embedder_only=False, return_cla
...
+                chunksize=1,
+            )
+
+            samples = []
+            datasets = []
+            for dataset, tensor_names, sample in tqdm(results, total=dict_batches_to_process):
+                datasets.append(dataset)
+                samples.extend(sample)
+
+            concat_datasets = ConcatDataset(datasets)
```
tholor: For large files it might blow up memory because we keep all "datasets" and all "samples" in memory. It might be an alternative to also call "_run_inference" on the chunks only and concatenate the resulting "preds"

PR #131, opened 2019-10-30
Refactor squad qa

Timoeller:
This PR includes
QA eval during training
working QA inference mode

missing: support for other models except Bert


farm/modeling/prediction_head.py
```diff
@@ -849,66 +881,174 @@ def prepare_labels(self, start_position, end_position, **kwargs):
...
+                pred["answers"] = passage_predictions
+                all_preds.append(pred)
 
         result["predictions"] = all_preds
         return result
 
-    def _get_best_indexes(self, logits, n_best_size):
-        """Get the n-best logits from a numpy array."""
-        idx = np.argsort(logits,axis=1)[:,-n_best_size:]
+    def _aggregate_predictions(self, preds):
```
tholor: Can we rename this to "aggregate_preds" to be more compliant with the rest (formatted_preds, logits_to_preds...). I would also make this a public method, since there might be cases (e.g. in Branden's branch in eval) where we call this method from the outside.

farm/modeling/prediction_head.py
```diff
@@ -849,66 +881,174 @@ def prepare_labels(self, start_position, end_position, **kwargs):
...
+
+    def _get_best_textanswer_indices(self, logits, n_best_size):
+        """Get the n-best logits from a numpy array without considering the zero index.
+        zero index corresponds to no answer, which we deal with separately"""
+        logits_without_zero = logits[:,1:]
+        idx_without_zero = np.argsort(logits_without_zero,axis=1)[:,-n_best_size:]
+        idx = idx_without_zero + 1
         return idx
+
+class SquadHead(QuestionAnsweringHead):
```
tholor: Why do we need this extra head? Can't we move this load() method into QuestionAnsweringHead itself?

farm/infer.py
```diff
@@ -189,42 +178,50 @@ def inference_from_dicts(self, dicts, rest_api_schema=False, use_multiprocessing
...
+                            preds_all.append(self._run_inference_qa(dataset, tensor_names, samples))
+                        else:
+                            preds_all.extend(self._run_inference(dataset, tensor_names, samples))
                         pbar.update(multiprocessing_chunk_size)
 
         else:
             chunk = next(grouper(dicts, len(dicts)))
-            dataset, tensor_names, sample = self._multiproc(chunk, processor=self.processor, rest_api_schema=rest_api_schema)
-            preds_all = self._run_inference(dataset, tensor_names, sample)
+            dataset, tensor_names, baskets = self._multiproc(chunk, processor=self.processor, rest_api_schema=rest_api_schema)
```
tholor: Why not getting samples back from _multiproc() instead of baskets (as we did before)? Would make the code here easier to read by moving the unpacking part to multiproc()

Issue #115, opened 2019-10-14
Please add `proxies=proxies` capability to any functions call to `requests`


cregouby:
**Is your feature request related to a problem or particular use case?**
As a corporate user, I need to pass through proxies for any https request. Currently, framework initialization commands are missing the `proxies=proxies` parameter pass-trough to the `requests` library, making initalization very painful with manual download and badly reproducible.

**Describe the solution you'd like**
Please incorporate `proxies=proxies` parameter to any call to the `requests` library so that initialisation like
```
proxies = {'https': config['https_proxy']}
tokenizer = BertTokenizer(
    vocab_file="bert-base-cased",
    proxies = proxies,
    do_lower_case=False)
```
is now OK through corporate proxies.

**Additional context**
I manage to workaround it through merging with [huggingface/transformers/.../file_utils.py](https://github.com/huggingface/transformers/blob/master/transformers/file_utils.py) web related functions...

Timoeller:
Can it be that this functionality is already in place? Can you double check @tholor please.

Did you, @cregouby  try to use the BertTokenizer like this? Does it produce any errors?

cregouby:
Yes, from the tutorial notebook, right now executing
```
tokenizer = BertTokenizer.from_pretrained(
    pretrained_model_name_or_path="bert-base-cased",
    do_lower_case=False)
```
gives
```
10/15/2019 09:11:52 - INFO - pytorch_transformers.file_utils -   https://s3.amazonaws.com/models.huggingface.co/bert/bert-base-cased-vocab.txt not found in cache, downloading to /tmp/tmpg0u0g9n0
10/15/2019 09:11:52 - ERROR - pytorch_transformers.tokenization_utils -   Couldn't reach server to download vocabulary.
```

Timoeller:
How about you adjust the code in our tutorial notebook to:
```
tokenizer = BertTokenizer.from_pretrained(
    pretrained_model_name_or_path="bert-base-cased",
    proxies = proxies,
    do_lower_case=False)
```
and set the proxies yourself?

tholor:
Supporting proxies is definitely an important feature since many corporates face this issue. We will have a look how to integrate it most smoothly into FARM. I see different options from a) passing the proxy arg all the way down to the requests library to b) reading a global environment variable. @tanaysoni can you please follow up on this?

tholor:
Thanks for the update @cregouby and the PR. 
I see 3 parts of the FARM code that rely on proxies:
[...]

PR #135, opened 2019-11-04
complete the fix to #115 with _download_extract_downstream_data() 

cregouby:


cregouby:
processor file_to_dicts need kwargs passthrough

PR #147, opened 2019-11-14
This should fix #146

### LINKED ISSUE #146: doc_classification_cola example uses wrong method for inference
In line 107 of `doc_classification_cola.py ` the method `run_inference` for Inferencer is used but this should really be `inference_from_dicts`

This is present in master as of commit  41644de7a0b8331b6d1ba1aeca9c77431509b4dc

johann-petrak:


tholor:
Great! Thanks for reporting and fixing @johann-petrak!

PR #155, opened 2019-11-20
Fix loading custom vocab in transformers style for LM finetuning

tholor:
LM finetuning with custom vocab was broken after switching to the transformer style of handling custom vocab (adding tokens instead of using "unused tokens"). 

The complication here is that with a larger vocab we need to adjust both the size of the embedding layer in the LM and the decoder (bias+weights) in the PH. In addition, the decoder shares the weights with the embedding layer.  

We therefore need to supply now an extra arg "n_added_tokens" for loading the PH / LM. 

Example: 
```
    ...
    tokenizer.add_tokens(["somecustomtoken", "specialrareword"])

    ...
    language_model = LanguageModel.load(lang_model, n_added_tokens=len(tokenizer.added_tokens_decoder))
    lm_prediction_head = BertLMHead.load(lang_model, n_added_tokens=len(tokenizer.added_tokens_decoder))
```

### DIFF
--- a/farm/modeling/language_model.py
+++ b/farm/modeling/language_model.py
@@ -109,7 +109,7 @@ class LanguageModel(nn.Module):
             model_emb_size = language_model.model.resize_token_embeddings(new_num_tokens=None).num_embeddings
             vocab_size = model_emb_size + n_added_tokens
             logger.info(
-                f"Tokenizer vocabulary (len: {vocab_size}) does not match original language model vocabulary (len: {model_emb_size}). Resizing embedding layer of LM accordingly")
+                f"Resizing embedding layer of LM from {model_emb_size} to {vocab_size} to cope for custom vocab.")
             language_model.model.resize_token_embeddings(vocab_size)
             # verify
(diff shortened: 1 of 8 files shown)


farm/modeling/language_model.py
```diff
@@ -103,6 +103,18 @@ def load(cls, pretrained_model_name_or_path, **kwargs):
                 f"models: https://farm.deepset.ai/api/modeling.html#farm.modeling.language_model.LanguageModel.load"
             )
 
+        # resize embeddings in case of custom vocab
+        if n_added_tokens != 0:
+            # TODO verify for other models than BERT
+            model_emb_size = language_model.model.resize_token_embeddings(new_num_tokens=None).num_embeddings
+            vocab_size = model_emb_size + n_added_tokens
+            logger.info(
+                f"Tokenizer vocabulary (len: {vocab_size}) does not match original language model vocabulary (len: {model_emb_size}). Resizing embedding layer of LM accordingly")
```
brandenchan: It would be good to rephrase this message if possible. We want to differentiate cases where the vocab size mismatch arises due to wrong components vs the introduction of added vocab

PR #156, opened 2019-11-20
Fix tokenizer for multiple whitespaces

tholor:
Text that contained multiple whitespaces different to " " (e.g. "\r\r", "\n\n" ...) caused weird behaviour in the transformers tokenizer leading to cases where "\r\r" was tokenized as "[SEP]". 

Added a simple fix for that and extended the test cases for such texts.


brandenchan:
Normalizing multi whitespace sequences is a good idea! Looks good

Issue #51, opened 2019-08-07
Change classification_report param in evaluation to generic report


busyxin:
Update eval.py to manage reports that are not a classification

tholor:
Fixed by #173

PR #164, opened 2019-11-28
Add more flexibility to eval metrics by allowing custom functions

### LINKED ISSUE #152: Make it easier to use own metrics, log more metrics
Currently the metric (for TextClassificationProcessor) needs to get specified as a string and picked from one of the pre-defined ones. That metric will then also get logged during training.

It would be very useful if it was possible to specify my own metric definition instead by allowing a function to get passed which takes preds/labels and returns the name and value of the metric. 

Also it would be very useful to specify a list of metrics so that several metrics can be logged (e.g. log the per-label f1 for classification)

tholor:
This PR fixes #152 and #51:

1) renaming the arg "classification_report" to a more general "report"
2) Allowing more flexible eval metrics: 
a) custom functions

```
def my_metric(preds, labels):
   # do whatever measure you like & return a numerical value
    return 123

processor = TextClassificationProcessor(...
                                        metric=my_metric,
                                            ..)
```
b) multiple metrics by providing a list:
```
processor = TextClassificationProcessor(...
                                        metric=[my_metric, "acc"],
                                            ..)
```

Limitations: 
1) Multiprocessing requires pickling of the custom metric function. Therefore it's important to define the function at the top level of the module (see details [here](https://stackoverflow.com/questions/8804830/python-multiprocessing-picklingerror-cant-pickle-type-function)) 
2) Saving / Loading of the processor becomes tricky, when a custom metric function is used because it's not serializable to json anymore. We would need to go for another format (e.g. pickling), to fully support this, but this is currently out of scope. Therefore, I implemented a workaround: we don't save the function but a placeholder string "custom_metric_fn". This will allow loading the processor as usual for inference. However, you won't be able to save the procesor, load it again and continue training / eval. This should be however a very rare use case.


tholor:
I actually like the design proposed by @johann-petrak in #165 better. Registering a custom metric function instead of passing it directly seems an acceptable workflow for the user and we avoid pickling in loading/saving the processor. I will work on bringing these two PRs together ...

PR #167, opened 2019-12-10
Issue59 - Add cross-validation for small datasets

johann-petrak:
This is just a rough idea of how xval could get implemented with almost only what is already there.

This basically uses two "tricks" to do the xval:
* implement a replacement data silo to provide the train/dev/eval sets for each fold based on the original data
* use the eval results to also store the predictions and labels for later so we can gather them to calculatee over-all-folds metrics after the xval. Make the eval module ignore those fields when logging, if we use a name that starts with underscore. 
* all the splitting into folds is done using the methods from sklearn 

This example is meant to do a full training run on each fold using early stopping and saves each per-fold best model in a separate directory.
The xval is carried out on the combined train and dev parts of the original dataset, from all that data, we generate k stratified folds of train+test, and then use a portion of that train as dev.
After the xval, the overal best model from all folds with regard to one of the metrics is used for evaluation on the original test set. 

Let me know what you think of this, especially:
* the idea to use fields with names that start with an underscore to pass on data like this which does not get logged
* the idea to do the xval using a lightweight data-silo replacement

### DIFF
--- a/examples/doc_classification_crossvalidation.py
+++ b/examples/doc_classification_crossvalidation.py
@@ -34,12 +34,12 @@ ml_logger = MLFlowLogger(tracking_uri="logs")
 ##########################
 ########## Settings
 ##########################
-xval_folds = 3 #10
+xval_folds = 5
 xval_stratified = True
 
 set_all_seeds(seed=42)
 device, n_gpu = initialize_device_settings(use_cuda=True)
-n_epochs = 1
+n_epochs = 20
 batch_size = 32
 evaluate_every = 100
 lang_model = "bert-base-german-cased"
--- a/examples/doc_classification_crossvalidation.py
+++ b/examples/doc_classification_crossvalidation.py
@@ -2,7 +2,7 @@
 import logging
 import json
 
-from farm.data_handler.data_silo import DataSilo, DataSilo4Xval
+from farm.data_handler.data_silo import DataSilo, DataSiloForCrossVal
 from farm.data_handler.processor import TextClassificationProcessor
 from farm.modeling.optimization import initialize_optimizer
 from farm.modeling.adaptive_model import AdaptiveModel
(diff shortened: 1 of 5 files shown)


tholor:
Thanks for the PR @johann-petrak! 
Happy to see an option to use crossvalidation in FARM. I think the overall workflow is decent and will help on small datasets.
Regarding the design: 
- Using a light weight datasilo wrapper for cross val is a pragmatic choice. I see some potential to simplify this, but this should be addressed in a future PR.
- Passing preds and labels via the metrics seems not ideal to me. It would always require the user to register custom metrics including these two even if he want to use a standard metric. I will have a look if there's another option and come up with a suggested implementation (e.g. storing preds and labels to disk).

johann-petrak:
Thanks for the feedback! I agree with your argument about having to define a metric to pass the predictions and targets. My suggestion would have been to change this so that instead the evaluation method gets a parameter that tells it to inject those values into the result  and I see you have implemented just that in your  commit! :)
However I still like the idea that one could add arbitrary additional fields that do not get logged to the result, so I would have stayed with using the hidden underscore names for pred and labels and left the feature to not log any such field. 

I fell ill yesterday so my brain may not work properly and I could not look in detail at all the changes you made.  One thing I noticed is that you changed the maximum number of epochs to one, why? I think that way the whole earlystopping becomes useless?

tholor:
>  I still like the idea that one could add arbitrary additional fields that do not get logged to the result [...]

Sure, why not. Added your check back into the code :) 

> you changed the maximum number of epochs to one [...]

My bad. Was only to speed up some debugging I did. Changed epochs back to 20.

Do you see anything that stops us from merging?

johann-petrak:
The only thing that may be useful to add is a short line to the documentation of Trainer(evaluator_test, evaluator_dev): the evaluator_test method handles the three cases None, Evaluator, False differently while the evaluator_dev supports Evaluator and None. 
Otherwise all looks great to me!

PR #170, opened 2019-12-12
Train from scratch

brandenchan:
This branch implements language model training from scratch


brandenchan:
Currently the models are not able to overfit to even small datasets

tholor:
This might be an issue due to parameter initialization or it's just the normal model behaviour in this combination of dataset, drop_out, low learning rate etc.
We should try to run the same training using google's original script and compare if that one actually overfits.

tholor:
Closing as this was already merged as a part of #188

PR #171, opened 2019-12-12
Encoding of QA IDs

brandenchan:
This implements a more general id encoding strategy that is not specific just to SQUAD. There  are also some general error reporting and print out improvements


farm/modeling/prediction_head.py
```diff
@@ -1166,6 +1164,8 @@ def reduce_preds(self, preds, n_best=5):
         returns the n_best predictions on the document level. """
 
         # Initialize some variables
+        # no_ans_threshold is how much greater the no_answer logit needs to be over the pos_answer in order to be chosen
+        no_ans_threshold = 0
         document_no_answer = True
```
Timoeller: Why do we have no answer threshold and a flag "document no answer"?

farm/utils.py
```diff
@@ -27,16 +27,16 @@ def set_all_seeds(seed, n_gpu=0):
...
+    # automatic adjustment of multiprocessing chunksize
+    # for small files (containing few dicts) we want small chunksize to ulitize all available cores but never less
+    # than 2, because we need it to sample another random sentence in LM finetuning
+    # for large files we want to minimize processor spawning without giving too much data to one process, so we
+    # clip it at 5k
+    multiprocessing_chunk_size = int(np.clip((np.ceil(dicts_per_cpu / 5)), a_min=min_chunksize, a_max=max_chunksize))
+    # This lets us avoid cases in lm_finetuning where a chunk only has a single doc and hence cannot pick
+    # a valid next sentence substitute from another document
+    while num_dicts % multiprocessing_chunk_size == 1:
+        multiprocessing_chunk_size -= -1
```
Timoeller: Why is this code in the PR? It seems as if you did not work on that?

farm/modeling/language_model.py
```diff
@@ -354,7 +354,10 @@ def load(cls, pretrained_model_name_or_path, language=None, **kwargs):
 
         """
         albert = cls()
-        albert.name = pretrained_model_name_or_path
+        if "farm_lm_name" in kwargs:
+            albert.name = kwargs["farm_lm_name"]
+        else:
+            albert.name = pretrained_model_name_or_path
```
Timoeller: I think its ok for now, but having unrelated changes in a PR is not best practice.

farm/data_handler/utils.py
```diff
@@ -288,6 +289,8 @@ def _get_random_sentence(all_baskets, forbidden_doc):
             rand_sent_idx = random.randrange(len(rand_doc))
             sentence = rand_doc[rand_sent_idx]
             break
+    if sentence is None:
+        raise Exception("Failed to pick out a suitable random substitute for next sentence")
```
Timoeller: This is changes about LM finetuning? Why is it in here?

PR #174, opened 2019-12-16
WIP Add option to limit number of processes in datasilo

tholor:
In some cases we might not utilize all available CPU cores for multiprocessing (e.g. debugging).
We should allow the user to set a limit of max_cpus

### DIFF
--- a/farm/data_handler/data_silo.py
+++ b/farm/data_handler/data_silo.py
@@ -132,7 +132,8 @@ class DataSilo:
                 results = map(partial(self._dataset_from_chunk, processor=self.processor), grouper(dicts, num_dicts))
 
             datasets = []
-            with tqdm(total=len(dicts), unit=" Dicts", desc="Converting Dicts to PyTorch Datasets") as pbar:
+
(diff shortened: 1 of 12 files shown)


farm/data_handler/data_silo.py
```diff
@@ -93,29 +101,42 @@ def _get_dataset(self, filename, dicts=None):
...
+                logger.info(
+                    f"Multiprocessing disabled, using a single worker to convert {num_dicts}"
+                    f"dictionaries to pytorch datasets."
+                )
+
+                results = map(partial(self._dataset_from_chunk, processor=self.processor), grouper(dicts, num_dicts))
 
             datasets = []
-            with tqdm(total=len(dicts), unit=' Dicts') as pbar:
+            with tqdm(total=len(dicts), unit=" Dicts", desc="Converting Dicts to PyTorch Datasets") as pbar:
```
tholor: Just added another description here earlier today: https://github.com/deepset-ai/FARM/pull/175/files
I would prefer "Preprocessing Dataset" here since dicts is not very clear for users who are not familiar with the FARM processors yet

PR #179, opened 2019-12-20
Fixed micro f1 score

maknotavailable:


tholor:
Thanks for fixing :+1:

PR #180, opened 2019-12-21
Fix for use AMP instead of naive fp16 (#133)

maknotavailable:
Was missing the full integration of the updated initialize_optimizer() here.


tholor:
Great! Thanks for spotting & fixing :+1:

PR #181, opened 2020-01-03
WIP: add XLM-Roberta - first shot

Timoeller:


examples/ner.py
```diff
@@ -85,7 +85,7 @@
 model = trainer.train(model)
 
 # 8. Hooray! You have a model. Store it:
-save_dir = "saved_models/bert-german-ner-tutorial"
```
tholor: Let's keep the old path here `saved_models/bert-german-ner-tutorial`

PR #183, opened 2020-01-06
Params in Question Answering Head

brandenchan:
The params n_best, context_window_size and no_ans_threshold can now be set when initializing QuestionAnsweringHead


brandenchan:
> Looking ok, though why did you remove the n_best changes again?

That was a mistake. I have recommitted them

PR #195, opened 2020-01-16
Automatically infer layer dims of prediction head

brandenchan:
Turn layer_dims into an optional argument. They can now be inferred from other sources. This means that the prediction head's  feed forward layer can be resized twice (once for input dimensions, once for output dimensions)


farm/modeling/prediction_head.py
```diff
@@ -151,6 +151,39 @@ def prepare_labels(self, **kwargs):
...
+        """ This function compares the number of labels against the output dimensionality
+        of the prediction head. If there is a mismatch, the prediction head will be resized to fit."""
+        if "feed_forward" not in dir(self):
+            return
+        else:
+            old_dims = self.feed_forward.layer_dims
+            if output_dim == old_dims[-1]:
+                return
+            new_dims = old_dims[:-1] + [output_dim]
+            logger.info(f"Resizing input dimensions of {type(self).__name__} ({self.task_name}) "
```
tholor: typo "Resizing output dimensions"

PR #198, opened 2020-01-17
Implement next_sent_pred flag

brandenchan:
Previously the next_sent_pred argument in the BertStyleLMProcessor was not working if the NextSentenceHead is passed to the adaptive model. Now the NextSentenceHead will be removed from the Adaptive model if next_sent_pred=True. This allows for turning off NSP by just setting the argument to False.


brandenchan:
This also works for the case where just one prediction head is passed into the Adaptive Model

PR #203, opened 2020-01-21
QA example

brandenchan:
The QA example script was broken in a few different places. It now runs now that inference and write_squad_predictions() have been fixed

### DIFF
--- a/examples/question_answering.py
+++ b/examples/question_answering.py
@@ -56,8 +56,8 @@ def question_answering():
         data_dir=Path("../data/squad20"),
     )
 
-
     # 3. Create a DataSilo that loads several datasets (train/dev/test), provides DataLoaders for them and calculates a few descriptive statistics of our datasets
+    # NOTE: In FARM, the dev set metrics differ from test set metrics in that they are calculated on a token level instead of a word level
     data_silo = DataSilo(processor=processor, batch_size=batch_size, distributed=False)
 
     # 4. Create an AdaptiveModel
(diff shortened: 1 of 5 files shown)


examples/question_answering.py
```diff
@@ -124,5 +124,10 @@ def question_answering():
         out_filename="predictions.json"
     )
 
+    # To get evaluate the model's performance on the SQuAD dev set, run the farm/squad_evaluation.py
```
tholor: Can you make the value / use case of this clearer? Why should they consider doing this and not just take the FARM metrics produced during training? -> "To get metrics that are comparable to the SQuAD leaderboard and most other frameworks, you can now run the "official" squad eval script, which produces word level metrics and handles the no_answer option slightly different"

farm/modeling/adaptive_model.py
```diff
@@ -271,6 +274,7 @@ def connect_heads_with_processor(self, tasks, require_labels=True):
             else:
                 num_labels = len(label_list)
             head.resize_output(num_labels)
+            head.to(self.device)
```
tholor: Wasn't that already part of a previous PR?

examples/question_answering.py
brandenchan: Yup, this comment has been improved in the latest commit

farm/modeling/adaptive_model.py
brandenchan: Yes this was not the cleanest way to introduce this change into the branch

PR #206, opened 2020-01-22
Undo Prediction Head Output Resizing 

brandenchan:
The way that prediction head output resizing was happening meant that the prediction head's weight matrix was at risk of not receiving gradient updates. This PR undoes those changes. Now users need to specify the number of labels when using the TextClassificationHeads or TokenClassificationHeads

### DIFF
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -240,7 +240,7 @@ class RegressionHead(PredictionHead):
 class TextClassificationHead(PredictionHead):
     def __init__(
         self,
-        layer_dims=[768,1],
+        layer_dims=None,
         num_labels=None,
         class_weights=None,
         loss_ignore_index=-100,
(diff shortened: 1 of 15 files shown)


farm/modeling/prediction_head.py
```diff
@@ -258,16 +241,32 @@ class TextClassificationHead(PredictionHead):
     def __init__(
         self,
         layer_dims=[768,1],
```
tholor: I think having these defaults can now become confusing (e.g. people not supplying any of layer_dims and num_labels). 
How about:
- layer_dims = None 
- in init: 
``` 
if num_labels: 
     self.layer_dims = [768, num_labels]   
elif layer_dims: 
     self.layer_dims = layer_dims                
      (and maybe raise a deprecation warning)
else: 
    raise ValueError("Please supply `num_labels` to define output dim of prediction head" )
```

brandenchan: Yup agreed. This is fixed in the latest commit.

PR #215, opened 2020-01-27
Ensure ALBERT and XLNET tokenisers keep accents

brandenchan:
The keep_accents argument of the ALBERT and XLNET tokenizers need to be explicitly set to True otherwise umlauts will be lost when processing German text.


brandenchan:
I looked through the arguments of the different tokenizers and only XLNet and ALBERT seem to have the keep_accents argument.

Issue #182, opened 2020-01-06
What multi-task strategies are implemented?


AndriyMulyar:
**Question**
Suppose I have three tasks A,B,C each with a corresponding head and loss. What multi-task training strategies are implemented?

tholor:
Hey @AndriyMulyar, 

We currently have hard parameter sharing and the simultaneous learning of multiple (auxiliary) tasks (similar to what Sebastian Ruder describes in https://arxiv.org/pdf/1706.05098.pdf). This means in your example the tasks A, B, C would share the parameters of the language model (e.g. BERT) and only the weights from the head are "task-specific". The individual loss of each head is summed up to a total loss and a single backprop is then used to update the weights for each batch. So the model is really learning all tasks at the same time. Changing the aggregation type (e.g. mean) or weighting the loss differently should be straight forward.

We made some good experiences using this type of multi-task learning to make a target task more robust by adding auxiliary tasks. For example:
- You want to classify A. Training on classifying A and B helps to improve on A.
- You want to train a QA model. Training on QA + MaskedLM can help to make QA more robust to domain shifts.

**Usage:**
From https://github.com/deepset-ai/FARM/releases/tag/0.2.2

1. Add the tasks to the Processor:
```
    processor = TextClassificationProcessor(...)

    news_categories = ["Sports", "Tech", "Politics", "Business", "Society"]
    publisher = ["cnn", "nytimes","wsj"]

    processor.add_task(name="category", label_list=news_categories, metric="acc", label_column_name="category_label")
    processor.add_task(name="publisher", label_list=publisher, metric="acc", label_column_name="publisher_label")
```

2. Link the data to right `PredictionHead` by supplying the task name at initialization:
``` 
category_head = MultiLabelTextClassificationHead(layer_dims=[768,5)], task_name="action_type")
publisher_head = MultiLabelTextClassificationHead(layer_dims=[768, 3], task_name="parts")
```

What type of multi-task strategy are you interested in?

AndriyMulyar:
Hi, thanks for the response! The obvious approach to me seems to be training each head in a round robin scheme. Are you familiar with any analysis on various strategies for multi-task training? For instance, it is not obvious (to me) how one would perform updates with loss averaging when tasks with different loss functions are present (say cross entropy for a token classification head and MSE for an STS head).

tholor:
[...]

PR #220, opened 2020-01-29
Add option for custom loss aggregation of heads

tholor:
**Context:** 
Each prediction head produces a loss that gets then aggregated to a full model loss. So far we used `sum()` as the aggregation function. 

**Problem:**
This is not optimal in the case of multiple PHs as it impacts the scale of the loss and therefore might require adjustments of the learning rate. 
It also doesn't allow weighting of different prediction heads, where one task might be more important or just on a different loss scale.

**Solution:**
I would suggest to make this more flexible and let the user define a custom strategy via 
`loss_aggregation_fn`, which can be passed when initializing the AdaptiveModel. 

Default of `loss_aggregation_fn` will be sum(), but you can configure any fn that [Tensor, Tensor ...] and returns a single Tensor.

**Example:**
``` 
    import torch
    loss_fn = lambda x: torch.mean(torch.stack(x), dim=0)

    model = AdaptiveModel(
        language_model=language_model,
        prediction_heads=[lm_prediction_head, next_sentence_head],
        embeds_dropout_prob=0.1,
        lm_output_types=["per_token", "per_sequence"],
        device=device,
        loss_aggregation_fn=loss_fn
    )
```
Related to discussion in #182 

What do you think @johann-petrak ? 
Would that also cover your use cases?

### DIFF
--- a/farm/modeling/adaptive_model.py
+++ b/farm/modeling/adaptive_model.py
@@ -44,7 +44,7 @@ class AdaptiveModel(nn.Module):
                                     Note: The loss at this stage is per sample,
                                     i.e one tensor of shape (batchsize) per prediction head.
                                     Default is sum(), but you can configure any fn that takes
-                                    [Tensor, Tensor ...] and returns [Tensor].
+                                    [Tensor, Tensor ...] and returns a single Tensor.
         :type loss_aggregation_fn: function
         """
 
--- a/farm/modeling/adaptive_model.py
+++ b/farm/modeling/adaptive_model.py
@@ -22,7 +22,8 @@ class AdaptiveModel(nn.Module):
         prediction_heads,
         embeds_dropout_prob,
         lm_output_types,
-        device
+        device,
+        loss_aggregation_fn=sum
     ):
         """
         :param language_model: Any model that turns token ids into vector representations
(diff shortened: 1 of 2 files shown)


johann-petrak:
I think this needs a bit more thinking through. I am not too sure myself, since what I mentioned in #182 is based on what I am planning to do, not what I already have experience with. 

My main concern is that the method for combining the loss may depend on the context, where the context could be global (in this case your solution would be sufficient because the lambda could be a closure over the global data), but also could be the concrete instances or the concrete batch. 

For example, the way how we want to combine the losses could depend on the targets for each head. 
Another completely different approach could depend on the training step, e.g. when we want to boost the importance of each head in a round robin fashion every kth batch. 
I have no idea which of these things are really useful or required, just thinking they *may* turn out to be important. 

So I guess ideally we would pass a couple of additional parameters to the function which allow the user to access all these things *if needed* or otherwise they can ignore those parameters.

johann-petrak:
That one looks good, all there I can think of now! :)

tholor:
Great :)
It's a bit of a trade-off here between giving flexibility and keeping the code clean and easy to understand. For now, this is the best compromise I could come up with. Let's merge it and reflect on it once we have a clearer understanding of the more complex use cases (e.g. loss depending on batches).

tholor:
@AndriyMulyar, this PR might be helpful for your work on MTL. While it doesn't implement a round-robin dataloader, you could basically switch on/off or weight the loss of prediction heads when you iterate over batches (batch 1 => only the loss of head 1, batch 2 = loss of head 2 ...)

Code sketch: 
```
def loss_aggregation_fn(tensors, global_step, batch=None):
    if global_step % 2:
        return tensors[0]
    else:
        return tensors[1]
...

model = AdaptiveModel(
        language_model=language_model,
        prediction_heads=[head1, head2],
        embeds_dropout_prob=0.1,
        lm_output_types=["per_sequence","per_sequence"],
        device=device,
        loss_aggregation_fn=loss_aggregation_fn)
```

PR #230, opened 2020-02-03
Fix shape of logits in regression head

### LINKED ISSUE #229: Using a target size ... that is different to the input size
Please what does this mean ?

Train epoch 1/20 (Cur. train loss: 218.7825):  61%|██████▏   | 62/101 [01:40<02:25,  3.72s/it]/home/ec2-user/anaconda3/envs/python3/lib/python3.6/site-packages/torch/nn/modules/loss.py:431: UserWarning: **Using a target size (torch.Size([32, 1])) that is different to the input size (torch.Size([32]))**. This will likely lead to incorrect results due to broadcasting. Please ensure they have the same size.

tholor:
Fixes #229

### DIFF
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -204,9 +204,8 @@ class RegressionHead(PredictionHead):
         return logits
 
     def logits_to_loss(self, logits, **kwargs):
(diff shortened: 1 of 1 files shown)


zolekode:
Yes the error is gone. Awesome

PR #236, opened 2020-02-05
Add ConnectionError handling for MLFlow logger

tanaysoni:
This PR adds error handling in the MLFlow Logger
- raise error and display error message(with guide to use MLFLow in local mode)when a remote MLFLow instance is unreachable
- catch exceptions when a MLFlow instance is unreachable when logging metrics, params, or artifacts. This prevents training from crashing when a MLFLow server is momentarily unavailable.

### DIFF
--- a/farm/utils.py
+++ b/farm/utils.py
@@ -8,6 +8,7 @@ import signal
 
 import numpy as np
 import torch
+from requests.exceptions import ConnectionError
 from torch import multiprocessing as mp
 import mlflow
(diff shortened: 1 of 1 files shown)


tanaysoni:
@Timoeller At other place like farm/experiment and the examples, the logging for MLFlow runs is done via the MLFlowLogger class, so the exception handling is covered. We should not need the additional try catch blocks.

PR #237, opened 2020-02-06
Add support for text pair classification (ASNQ) and ranking (MSMarco)

brandenchan:
This pull request adds support text pair classification where the input is two separate sentences and the output is some form of classification. In the case of ASNQ, the prediction states whether a question can be answered by a certain passage.


brandenchan:
Column and feature naming is still a bit messy (tokens and tokens_b). Also still need to assess whether the TextPairClassificationProcessor should be a stand alone class or be incorporated into TextClassificationProcessor.

brandenchan:
There are examples now for MSMarco and ASNQ (binary). There is an experiment config, only for ASNQ since MSMarco requires a lot of preprocessing and separate metric scripts.

examples/passage_ranking.py
```diff
@@ -0,0 +1,141 @@
...
+    device, n_gpu = initialize_device_settings(use_cuda=True)
+    n_epochs = 2
+    batch_size = 64
+    evaluate_every = 500
+    lang_model = "bert-base-cased"
+    label_list = ["0", "1"]
+    train_filename = "train.tsv"
+    dev_filename = "dev.tsv"
+    generate_data = True
+    data_dir = Path("../data/msmarco_passage")
```
tholor: Maybe add a link here to the dataset that users should download

examples/passage_ranking.py
```diff
@@ -0,0 +1,141 @@
...
+    predictions_filename = "predictions.txt"
+    train_source_filename = "triples.train.small.tsv"
+    qrels_filename = "qrels.dev.tsv"
+    queries_filename = "queries.dev.tsv"
+    passages_filename = "collection.tsv"
+    top1000_filename = "top1000.dev"
+
+
+    # 0. Preprocess and save MSMarco data in a format that can be ingested by FARM models. Only needs to be done once!
+    if generate_data:
```
tholor: Would it make sense to give here a quick example of the target format for the TextPairClassificationProcessor? (e.g text_a      text_b     label). Might help users to use it for their own dataset (which is not msmarco)

farm/data_handler/input_features.py
```diff
@@ -47,7 +47,7 @@ def sample_to_features_text(
         tokens_b,
         add_special_tokens=True,
         max_length=max_seq_len,
-        truncation_strategy='do_not_truncate' # We've already truncated our tokens before
+        truncation_strategy='only_second'
```
tholor: Why do we need this change here?

farm/data_handler/processor.py
```diff
@@ -491,6 +493,32 @@ def _sample_to_features(self, sample) -> dict:
         )
         return features
 
+class TextPairClassificationProcessor(TextClassificationProcessor):
```
tholor: Please add a short docstring

farm/data_handler/processor.py
```diff
@@ -1069,6 +1099,7 @@ def __init__(
         self.delimiter = delimiter
         self.quote_char = quote_char
         self.skiprows = skiprows
+        self.label_remapping = label_remapping
```
tholor: What is this one doing here? If we need it, we should add a docstring

farm/data_handler/processor.py
tholor: It should be clear what input format is expected (e.g. colum names = "text"; "text_b", "label")

farm/evaluation/msmarco_passage_farm.py
```diff
@@ -0,0 +1,85 @@
+from farm.evaluation.msmarco_passage_official import compute_metrics_from_files
+import os
+import pandas as pd
+
+
+def msmarco_evaluation(preds_file, dev_file, qrels_file, output_file):
+    """
```
tholor: One line describing what type of evaluation we are doing here (e.g. with link to task website)

tholor: Also make the relation to the official script clearer. As I understood it, this is a wrapper to bring the predictions into the right shape and then run the official eval function "compute_metrics_from_files"?

farm/data_handler/processor.py
```diff
@@ -491,6 +493,32 @@ def _sample_to_features(self, sample) -> dict:
...
+            rename_columns=column_mapping,
+            filename=file,
+            delimiter=self.delimiter,
+            skiprows=self.skiprows,
+            proxies=self.proxies,
+        )
+        return dicts
+
+    def _dict_to_samples(self, dictionary: dict, **kwargs) -> [Sample]:
+        tokenized_question = tokenize_with_metadata(dictionary["text"], self.tokenizer)
```
tholor: I would keep the naming here generic, e.g. "tokenized_a" and tokenized_b

Issue #242, opened 2020-02-12
Allow efficient inference of multiple questions on one doc using REST API Format 


tholor:
**Is your feature request related to a problem or particular use case?**
QA with multiple questions for one doc. In contrast to other implementations, we should in this case only tokenize the doc **once** and reuse the results for each doc-question-pair. 

**Describe the solution you'd like**
We already have this functionality during Training and for Inference with the SQuAD input format: 
``` 
nlp = Inferencer.load("deepset/bert-base-cased-squad2, task_type="question_answering", gpu=True, max_seq_len=256, batch_size=200)
QA_input = [
    {
        "qas": [{
            "question": "What is Alphabet?",
            "id": 1,
            "answers": []},
            {"question": "What is driving innovations?",
            "id": 2,
            "answers": []}
],
        "context":  page
    }]
result = nlp.inference_from_dicts(dicts=QA_input, rest_api_schema=False)
```

However, this doesn't work yet for the simpler REST API input format:
``` 
QA_input = [{"questions": ["What is Alphabet?", "How about a second one?"],
             "text": "Some long text ..."}]
result = nlp.inference_from_dicts(dicts=QA_input, rest_api_schema=True)
``` 
throws 
``` 
Traceback (most recent call last):
  File "/home/ubuntu/miniconda3/envs/py37/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/home/ubuntu/pycharm/FARM/farm/infer.py", line 301, in _create_datasets_chunkwise
    dataset, tensor_names, baskets = processor.dataset_from_dicts(dicts, indices, rest_api_schema, return_baskets=True)
  File "/home/ubuntu/pycharm/FARM/farm/data_handler/processor.py", line 891, in dataset_from_dicts
    dicts = [self._convert_rest_api_dict(x) for x in dicts]
  File "/home/ubuntu/pycharm/FARM/farm/data_handler/processor.py", line 891, in <listcomp>
    dicts = [self._convert_rest_api_dict(x) for x in dicts]
  File "/home/ubuntu/pycharm/FARM/farm/data_handler/processor.py", line 965, in _convert_rest_api_dict
    raise ValueError("Inferencer currently does not support answering multiple questions on a text."
ValueError: Inferencer currently does not support answering multiple questions on a text.As a workaround, multiple input dicts with text and question pairs can be supplied in a single API request.
``` 
**Additional context**:
[...]

PR #246, opened 2020-02-13
Allow for multiple queries in QA inference when using rest_api format

brandenchan:
This PR allows for QA inference to handle multiple queries for a single doc when using rest api mode. This addresses #242

### DIFF
--- a/farm/infer.py
+++ b/farm/infer.py
@@ -252,11 +252,9 @@ class Inferencer:
             )
 
         # Assign a sequential doc_id to each dictionary
-        old_dicts = dicts
-        dicts = []
-        for i, d in enumerate(old_dicts):
-            d["document_id"] = i
-            dicts.append(d)
+        if rest_api_schema:
+            for i, d in enumerate(dicts):
+                d["document_id"] = i
 
         # Using multiprocessing
         if max_processes > 1:  # use multiprocessing if max_processes > 1
--- a/farm/infer.py
+++ b/farm/infer.py
@@ -251,11 +251,6 @@ class Inferencer:
                 f"b) ... run inference on a downstream task: make sure your model path {self.name} contains a saved prediction head"
             )
 
-        # Assign a sequential doc_id to each dictionary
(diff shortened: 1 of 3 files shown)


farm/infer.py
```diff
@@ -251,6 +251,13 @@ def inference_from_dicts(self, dicts, rest_api_schema=False, max_processes=128):
                 f"b) ... run inference on a downstream task: make sure your model path {self.name} contains a saved prediction head"
             )
 
+        # Assign a sequential doc_id to each dictionary
```
tholor: Shouldn't this only be executed for rest_api_schema=true? For SQuAD input format we want to use the passed doc id.

farm/infer.py
```diff
@@ -251,6 +251,13 @@ def inference_from_dicts(self, dicts, rest_api_schema=False, max_processes=128):
...
+        old_dicts = dicts
```
tholor: Maybe I am missing something here, but why do you need the "old_dict" var here? Can't we modify the dicts in place?

farm/infer.py
brandenchan: Actually there's a better way to do this now which is that the doc id for squad or rest api format is inferred from basket.id

farm/infer.py
brandenchan: You're right. Change made.

farm/infer.py
```diff
@@ -251,6 +251,11 @@ def inference_from_dicts(self, dicts, rest_api_schema=False, max_processes=128):
                 f"b) ... run inference on a downstream task: make sure your model path {self.name} contains a saved prediction head"
             )
 
+        # Assign a sequential doc_id to each dictionary
+        if rest_api_schema:
```
tholor: As I understood, we are now using baske_id instead. So we probably don't need this one anymore?

brandenchan:
Code was tested and now document ids are part of the output when rest_api_schema=True

PR #258, opened 2020-02-20
Refactor no answer handling in Question Answering

Timoeller:
This PR will need adjustments from downstream applications consuming QA inference, since we now return None for no answer (instead of empty string)

1. Change naming to no answer boost.
2. Change how no answer score is computed. 
3. Predictions contain a no answer along positive answers, the no answer score is relative to positive answers.

### DIFF
--- a/examples/question_answering.py
+++ b/examples/question_answering.py
@@ -36,8 +36,8 @@ def question_answering():
     evaluate_every = 2000
     lang_model = "roberta-base"
     do_lower_case = False # roberta is a cased model
-    train_filename = "train-sample2.json"#"train-v2.0.json"
-    dev_filename = "train-sample2.json" #"dev-v2.0.json"
+    train_filename = "train-v2.0.json"
+    dev_filename = "dev-v2.0.json"
 
     # 1.Create a tokenizer
     tokenizer = Tokenizer.load(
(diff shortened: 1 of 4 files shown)


examples/question_answering.py
```diff
@@ -36,8 +36,8 @@ def question_answering():
     evaluate_every = 2000
     lang_model = "roberta-base"
     do_lower_case = False # roberta is a cased model
-    train_filename = "train-v2.0.json"
-    dev_filename = "dev-v2.0.json"
+    train_filename = "train-sample2.json"#"train-v2.0.json"
```
tholor: back to original file names

farm/modeling/prediction_head.py
```diff
@@ -1103,14 +1104,14 @@ def formatted_preds(self, logits, baskets, rest_api_schema=False):
         preds_d = self.aggregate_preds(preds_p, passage_start_t, ids, seq_2_start_t)
         assert len(preds_d) == len(baskets)
 
-        # Separate top_preds list from the no_ans_gap float
-        top_preds, no_ans_gaps = zip(*preds_d)
+        # Separate top_preds list from the change_boost float. Add change_boost to current boost for switching
+        top_preds, change_boosts = zip(*preds_d)
```
tholor: I was first confused by the naming of "change_boosts" here. I would vote for keeping "no_ans_gap" or something like "no_ans_diff"

PR #278, opened 2020-03-13
Improve inferencer for better multiprocessing with QA / haystack

tholor:
- Adding arg to disable tqdm (is very messy with multiprocessing)
- Adding arg to set doc_stride for QA when loading an inferencer 
- Replacing ExitStack() block with pool.close() / pool.join() as it caused trouble with fastAPI / uvicorn deployments
- Doing one matrix operation for QA aggregation directly on GPU

### DIFF
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -1044,7 +1044,7 @@ class QuestionAnsweringHead(PredictionHead):
 
         # disqualify answers where end < start
         # (set the lower triangular matrix to low value, excluding diagonal)
-        indices = torch.tril_indices(max_seq_len, max_seq_len, offset=-1, device="cuda")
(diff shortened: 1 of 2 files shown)


farm/modeling/prediction_head.py
```diff
@@ -1044,7 +1044,7 @@ def logits_to_preds(self, logits, padding_mask, start_of_word, seq_2_start_t, ma
 
         # disqualify answers where end < start
         # (set the lower triangular matrix to low value, excluding diagonal)
-        indices = torch.tril_indices(max_seq_len, max_seq_len, offset=-1)
+        indices = torch.tril_indices(max_seq_len, max_seq_len, offset=-1, device="cuda")
```
tanaysoni: Would this work for a CPU? Maybe we could add the device as a property of PredictionHead.

tholor: Yep, good point :+1:

PR #279, opened 2020-03-16
Fix Train Step calculations for Checkpointing

tanaysoni:
Currently, the checkpointing during training is implemented at the beginning of the train loop. This causes a training step overlap. For instance, consider a case when `checkpoint_every` is set to 100 and a checkpoint is saved at step 300. Then, a training resumed from the saved checkpoint will start at step 300, causing step 300 to be executed twice.

This PR moves the checkpoint at the end of the training loop. Thus, when the training resumes from a checkpoint saved at step 300, it'd start from step 301.

### DIFF
--- a/farm/train.py
+++ b/farm/train.py
@@ -123,6 +123,7 @@ class Trainer:
         checkpoints_to_keep=3,
         from_epoch=0,
         from_step=0,
+        global_step=0,
     ):
         """
         :param optimizer: An optimizer object that determines the learning strategy to be used during training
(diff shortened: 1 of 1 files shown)


farm/train.py
```diff
@@ -464,7 +467,7 @@ def _save(self):
             pickle_module=dill,
         )
 
-        checkpoint_name = f"epoch_{self.from_epoch}_step_{self.from_step}"
+        checkpoint_name = f"epoch_{self.from_epoch + 1}_step_{self.from_step}"
```
tanaysoni: This is only to make the checkpoint folder names more readable. It doesn't affect the loading/saving behaviour of checkpoints.

tholor: Hm... so we store a checkpoint as "epoch 1" but need to load it via from_epoch=0? This sounds confusing to me.

PR #290, opened 2020-03-23
Add conversion of lm_finetuned to HF transformers

brandenchan:
This allows for models that have been lm finetuned in FARM to be converted to a HuggingFace Transformers model

### DIFF
--- a/farm/modeling/adaptive_model.py
+++ b/farm/modeling/adaptive_model.py
@@ -385,6 +385,8 @@ class AdaptiveModel(nn.Module):
             ph_state_dict["transform.LayerNorm.weight"] = ph_state_dict.pop("LayerNorm.weight")
             ph_state_dict["transform.LayerNorm.bias"] = ph_state_dict.pop("LayerNorm.bias")
             transformers_model.cls.predictions.load_state_dict(ph_state_dict)
+            logger.warning("Currently only the Masked Language Modeling component of the prediction head is converted, "
+                           "not the Next Sentence Prediction or Sentence Order Prediction components")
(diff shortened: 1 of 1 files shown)


farm/modeling/adaptive_model.py
```diff
@@ -374,6 +374,18 @@ def convert_to_transformers(self):
...
+            # init model
+            transformers_model = AutoModelWithLMHead.from_config(self.language_model.model.config)
+            # transfer weights for language model + prediction head
+            setattr(transformers_model, transformers_model.base_model_prefix, self.language_model.model)
+            ph_state_dict = self.prediction_heads[0].state_dict()
+            ph_state_dict["transform.dense.weight"] = ph_state_dict.pop("dense.weight")
+            ph_state_dict["transform.dense.bias"] = ph_state_dict.pop("dense.bias")
+            ph_state_dict["transform.LayerNorm.weight"] = ph_state_dict.pop("LayerNorm.weight")
+            ph_state_dict["transform.LayerNorm.bias"] = ph_state_dict.pop("LayerNorm.bias")
+            transformers_model.cls.predictions.load_state_dict(ph_state_dict)
```
tholor: I think you are converting here to a model with an MLM head but no NSP. `AutoModelForPreTraining` would give you a transformers model with both. I think for now we can keep it like this, as NSP is not that popular anymore these days and the FARM models might also only have the MLM head. However, please log an info message here like: "Converted to Transformer's 'AutoModelWithLMHead' with an MLM head only. No next sentence prediction head!"

PR #303, opened 2020-03-31
Fix ONNX device assertion for GPU and update session optimizations

tanaysoni:
This PR address the following:

- fix for device assertion when loading an ONNX model on a GPU
- remove the creation of an optimization graph during session creation as it's only meant for debugging
- add open MP optimization for inference on CPU

### DIFF
--- a/farm/modeling/adaptive_model.py
+++ b/farm/modeling/adaptive_model.py
@@ -4,6 +4,7 @@ import logging
 import os
 from pathlib import Path
 
+import psutil
 import numpy
 import onnxruntime
 import torch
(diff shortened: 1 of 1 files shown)


farm/modeling/adaptive_model.py
```diff
@@ -690,7 +691,7 @@ class ONNXAdaptiveModel(BaseAdaptiveModel):
     For inference, this class is compatible with the FARM Inferencer.
     """
     def __init__(self, onnx_session, prediction_heads, language, device="cpu"):
-        if str(onnxruntime.get_device()).lower() != str(device).lower():
+        if device == "cuda" and onnxruntime.get_device() != "gpu":
```
tholor: should device here really be a string or rather a `torch.device` as returned by `intialize_device_settings()`?

tanaysoni: Yes, it is intended to be `torch.device`. Also, the assertion is now simplified and the `device` argument is compliant with that of `AdaptiveModel`.

PR #304, opened 2020-03-31
Raise descriptive error in LanguageModel load if model is not found

tanaysoni:


farm/modeling/language_model.py
```diff
@@ -128,7 +128,11 @@ def load(cls, pretrained_model_name_or_path, n_added_tokens=0, language_model_cl
                 elif 'xlnet' in pretrained_model_name_or_path:
                     language_model_class = 'XLNet'
 
-            language_model = cls.subclasses[language_model_class].load(pretrained_model_name_or_path, **kwargs)
```
tanaysoni: Currently, if a model is not found, `LanguageModel.load()` fails with an unintuitive KeyError(trying to load `LanguageModel` subclasses with `language_model_class` as `None`).

tholor: Shouldn't we add a different Exception here, e.g. 
```
raise Exception(f"Could not determine the class of language model from the name {pretrained_model_name_or_path}. The model type must either be a part of the name (e.g. "xxx-bert-xxx") or specified manually via LanguageModel.load(..., language_model_class="Bert")") 
```

tanaysoni: The exception is outside the if-else block `if os.path.exists(config_file):`. So, when raising the exception we are not certain if it's due to config file not present at the supplied path or language model class is not determined.

tholor: Ok, but the exception we raise now in line 141 is misleading. I would then rather add an additional "else" after all the "elif" statements and raise the new exception there.

Issue #312, opened 2020-04-02
Verbose warnings for empty bucket


tholor:
**Describe the bug**
Seems like we introduced quite some verbose warnings with #308. 
@brandenchan can you please have a look and verify that we only print this warning if we actually removed some baskets without samples? It seems that we now always log this (once per worker).

**Message**
```
04/02/2020 12:34:22 - WARNING - farm.data_handler.processor -   Baskets with the following ids have been removed because they have no Samples: []
04/02/2020 12:34:22 - WARNING - farm.data_handler.processor -   Baskets with the following ids have been removed because they have no Samples: []
Preprocessing Dataset ../data/germeval18/train.tsv:   3%|█████████▏                                                                                                                                                                                                                                                                                                                      | 144/5009 [00:00<00:08, 555.22 Dicts/s]04/02/2020 12:34:22 - WARNING - farm.data_handler.processor -   Baskets with the following ids have been removed because they have no Samples: []
04/02/2020 12:34:22 - WARNING - farm.data_handler.processor -   Baskets with the following ids have been removed because they have no Samples: []
04/02/2020 12:34:22 - WARNING - farm.data_handler.processor -   Baskets with the following ids have been removed because they have no Samples: []
04/02/2020 12:34:22 - WARNING - farm.data_handler.processor -   Baskets with the following ids have been removed because they have no Samples: []
```

**To Reproduce**
run example/doc_classification.py

**System:**
 - OS: Ubuntu 18.04 
 - GPU/CPU: CPU
 - FARM version: current master

tholor:
Fixed by #313

PR #313, opened 2020-04-02
Less verbosity, Fix which Samples and Baskets being Thrown Away

brandenchan:
This PR removes a warning about removing baskets even when no baskets are thrown away.

This also fixes an issue where a whole Basket was thrown away if just one of its Samples was thrown away


brandenchan:
This addresses #312

PR #315, opened 2020-04-06
Add streaming mode for Inferencer

tanaysoni:
This PR adds a high-performance non-blocking streaming mode for large scale inference use cases. With this mode, the `dicts` parameter in `Inferencer.inference_from_dicts()` can optionally be a Python generator object that yields dicts, thus avoiding loading all dicts in memory. The `inference_from_dicts()` method would return a generator that yield inference predictions.

### DIFF
--- a/examples/streaming_inference.py
+++ b/examples/streaming_inference.py
@@ -21,7 +21,7 @@ def streaming_inference_example():
 
 def sample_dicts_generator():
     """
-    This is a sample dicts generator. Some example real world use-cases
+    This is a sample dicts generator. Some exemplary use cases:
 
     * read chunks of text from large files iteratively and generate inference predictions
     * connect with external datasources, eg, a Elasticsearch Scroll API that reads all documents from a given index
--- a/examples/streaming_inference.py
+++ b/examples/streaming_inference.py
@@ -1,0 +1,48 @@
+from farm.infer import Inferencer
+
+
+def streaming_inference_example():
+    """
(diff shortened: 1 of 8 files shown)


examples/streaming_inference.py
```diff
@@ -0,0 +1,48 @@
...
+    dicts = sample_dicts_generator()  # it can be a list of dicts or a generator object
+    results = inferencer.inference_from_dicts(dicts, num_processes=8, streaming=True, multiprocessing_chunksize=20)
+
+    for prediction in results:  # results is a generator object that yields predictions
+        print(prediction)
+
+
+def sample_dicts_generator():
+    """
+    This is a sample dicts generator. Some example real world use-cases
```
tholor, line 24: "Some exemplary use cases: "

farm/infer.py
```diff
@@ -225,95 +225,201 @@ def save(self, path):
...
+        * Classification / NER / embeddings: [{"text": "Some input text"}]
+
+
+        Inferencer has a high performance non-blocking streaming mode for large scale inference use cases. With this
+        mode, the dicts parameter can optionally be a Python generator object that yield dicts, thus avoiding loading
+        dicts in memory. The inference_from_dicts() method returns a generator that yield predictions. To use streaming,
+        set the streaming param to True and determine optimal multiprocessing_chunksize by performing speed benchmarks.
 
 
         :param dicts: Samples to run inference on provided as a list of dicts. One dict per sample.
```
tholor, line 281: "provided as a list of dicts (or a generator of dicts if streaming=True)"

Issue #325, opened 2020-04-16
Allow registration for custom eval report metrics


gsarti:
Dear maintainers,

I am fiddling with FARM to make it work for the very specific use case of multitask token-level regression with initial masking (as for NER), and in doing so I created my custom `TokenRegressionProcessor` and `TokenRegressionHead` classes. Alongside those I implemented various other utility methods, including a custom MSE metric that i register through `register_metric` in `metrics.py` to average the MSE across masked sentences with variable length in evaluation.

Everything works, except that if I want to evaluate the model during the training, there is no way to set the `report` parameter of the `Evaluator` instantiated in the `Trainer` class to `False` from the training script (I'm using the ones in `examples` as references). Currently available reporting functions won't support my use case (same as MSE above), so this makes the `Trainer` and `Evaluator` classes unusable for me, forcing me to implement them locally and either change the `report` to false, or manually add a custom token-level `r2` function in the reporting part of the `Evaluator`.

A solution would be to simply implement a registering system similar to the `register_metric` mentioned above also for reporting functions, instead of systematically raise `NotImplementedError` (line 118, `eval.py`). 

If you believe this is interesting, I will create a pull request. When my implementation of token-level regression will be complete I will gladly share my custom classes and methods if you belive if they may be worth including in your library.

Thanks in advance,

Gabriele Sarti

tholor:
Hey @gsarti,

Sounds like an interesting task you are working on. Just out of curiosity: For what use case do you do multitask token-level regression? 

Regarding the issue: 
I totally agree, that it would make sense adding an option to:
- disable reports
- register custom report_fn

For the first one, we could add a new arg to the `Trainer` that is passed down to the `Evaluator`. 
For the second one, I think a pattern similar to register_metric would do the job. 

It would be awesome if you could create a PR for that. We are of course happy to support you along the way, in case it's needed. 

Similarly, it would be great to get your `TokenRegressionProcessor` and `TokenRegressionHead` into FARM, once you have finished this :)

PR #331, opened 2020-04-20
Registration and parametrization for evaluation reports

gsarti:
Related to issue #325 that was deemed as interesting by maintainers:
- Moved code to compute report metrics inside `metrics.py`, refactoring it into the ad-hoc function `compute_report_metrics`.
- Added the function `register_ph_output_type` inside `metrics.py` to allow for custom metric registration in the context of evaluation reports.
- Added `eval_report` parameter in the `Trainer` class to allow switching off reporting when `evaluate_every` param is > 0.

### DIFF
--- a/farm/evaluation/metrics.py
+++ b/farm/evaluation/metrics.py
@@ -3,15 +3,28 @@ import numpy as np
 import pandas as pd
 from scipy.stats import pearsonr, spearmanr
 from seqeval.metrics import f1_score as ner_f1_score
-from sklearn.metrics import matthews_corrcoef, recall_score, precision_score, f1_score, mean_squared_error, r2_score
+from seqeval.metrics import classification_report as token_classification_report
+from sklearn.metrics import (
+    matthews_corrcoef,
+    recall_score,
+    precision_score,
+    f1_score,
+    mean_squared_error,
+    r2_score,
+    classification_report
+)
 from farm.utils import flatten_list
 import logging
 
 logger = logging.getLogger(__name__)
 
 registered_metrics = {}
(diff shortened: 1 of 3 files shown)


tholor:
Hey @gsarti , 

Thanks for working on this! I like the changes and would just propose some smaller adjustments (see my commit https://github.com/deepset-ai/FARM/pull/331/commits/536f1c38ae55141487a0c808e9d73e43233db0b4): 
- Renaming the method from `register_ph_output_type` to `register_report` (I totally see your reasoning for the naming, but I think it might confuse users. I added a docstring to clarify the interplay with `ph_output_type`)
- Allowing to use registered reports for existing prediction heads by checking **first** if there's a registered report

What do you think about these changes? Happy to merge this, when you are fine with it.

farm/evaluation/metrics.py
```diff
@@ -87,18 +102,19 @@ def compute_metrics(metric, preds, labels):
 
 
 def compute_report_metrics(head, preds, labels):
-    if head.ph_output_type == "per_token":
+    if head.ph_output_type in registered_reports:
+        report_fn = register_report[head.ph_output_type]
```
gsarti, line 106: I believe there is a typo here, shouldn't this be `registered_reports[head.ph_output_type]` instead of `register_report`? I think the bug was present in my version, too!

gsarti: Aside from that, looks good to me and you can definitely merge when it's fixed! Thanks again!

tholor: Oh yeah, good catch. I will fix it & merge.

PR #337, opened 2020-04-28
Implicitly connect heads with processor + check for connection

bogdankostic:


farm/modeling/adaptive_model.py
```diff
@@ -292,6 +297,8 @@ def logits_to_loss_per_head(self, logits, **kwargs):
         """
         all_losses = []
         for head, logits_for_one_head in zip(self.prediction_heads, logits):
+            # check if PredictionHead connected to Processor
+            assert hasattr(head, "label_tensor_name"), "PredictionHead has to be connected to Processor"
```
Timoeller, line 301: Lets create a more informative and less strict error message, since it might be the missing tensor names do not come from a missing connection between head and processor. 
I would write: f"Label_tensor_names are missing inside the {head.task_name} Prediction Head. Did you connect the model with the processor through either 'model.connect_heads_with_processor(processor.tasks)' or by passing the processor to the Adaptive Model?"

Issue #332, opened 2020-04-21
Accessing dataset index after evaluation


malteos:
**Question**

Hi all,

thanks for your great work! I was wondering why the dataset samples don't have an ID set. See https://github.com/deepset-ai/FARM/blob/89826c1f4269a5c0d0b5a6193ece94a3401766d9/farm/data_handler/processor.py#L492

Is there any way to obtain the ID/index/line number of a particular sample in the results of `Inferencer` or `Evaluator`? 

Or is it ensured that the order of the results is the same as in the test file, i.e. index of `results` list is equal to line number of `test.tsv`?

Best,
Malte

**Additional context**
-

Timoeller:
Hey @malteos thanks for the kudos : )

Your ID issue addresses a valid concern.
Short answer: You can disable multiprocessing by setting max_processes=1 in the data_silo, then the order of results is ensured. 

Long answer: We want to fill this ID value with sensible values and are working on a solution that works for all downstream tasks. In particular, we are not happy with the format of results in our Inferencer (nested dicts) and will transition to interface classes with defined attributes - one of them will be the ID. 
We cannot really commit to a timeline on this feature, though I hope the disabling of multiprocessing is a valid solution for you?

malteos:
Can I set a different `max_processes` for train and test? Processing all training data with a single process would be a time killer.. but just test set would be fine ;)

Timoeller:
I thought you wanted to use the Inferencer to get model predictions from a model that has been trained before (the Inferencer needs a trained model anyways)? 

So you could use multiprocessing during training for creating and saving the model.
Then you can load it without multiprocessing, so the order of results is assured with:
`model = Inferencer.load("path/to/FARM_model",num_processes=0)`
//you might have to update FARM, the num_processes inside the load() method is only a week old

You can do inference from dicts (with the structure from the example script) or inference from file, if you already divided train and test set.

brandenchan:
So we just tested out the inferencer even with multiprocessing and found that the output is in fact in the same order as the input. This is because we are calling self.process_pool.imap() and not imap_unordered() (see Inferencer._inference_with_multiprocessing()).

As for why the Sample ids are not set, this is because the id is actually assigned after TextClassificationProcessor.dict_to_samples. If you're interested in how this is done, have a look at  Processor._init_samples_in_baskets().

[...]

PR #342, opened 2020-04-30
Better access to test results (WIP)

malteos:
WIP regarding https://github.com/deepset-ai/FARM/issues/332

- Optional multi processing for test dataset (map results back to original IDs)
- Optional evaluation after training
- Reports as dict (not str) and include logits

(I didn't look into unit tests so it might crash something at the current status)

### DIFF
--- a/farm/data_handler/data_silo.py
+++ b/farm/data_handler/data_silo.py
@@ -44,6 +44,8 @@ class DataSilo:
         automatic_loading=True,
         max_multiprocessing_chunksize=2000,
         max_processes=128,
+        max_multiprocessing_chunksize_test=None,
+        max_processes_test=None,
         caching=False,
         cache_path=Path("cache/data_silo"),
     ):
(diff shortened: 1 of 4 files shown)


Timoeller:
Hey @malteos  thanks for the PR. Great to see people tweaking FARM! :dancers: 

**My general remark**: 
We would like to keep the Evaluator efficient and only produce eval metrics. Collecting all logits e.g. for the SQuAD dev set, would be quite an overhead.
If you want to output more stuff (like logits), we should bake this into the Inferencer (e.g. [here](https://github.com/deepset-ai/FARM/blob/master/farm/infer.py#L467)) and make the collection of logits optional in any case.
Btw. what do you need the logits for? Maybe there is a different way to work with them on your use case (in a PredictionHead)...

**Individual points:**
- Optional evaluation after training -> we have added this feature very recently in #335 but for your use case I believe it is sufficient to not load the test evaluator by setting test_filename=None in the Processor. (We should add documentation about this feature though.)
- Reports as dict (not str). That is a good idea, so people can access the results more individually. I think your intention was to access the logits though - right?

Timoeller:
Hey @malteos I wanted to follow up on this PR.

I think especially the question: "what do you need the logits for?" could be interesting to work on together. Maybe there is a different way to work with the logits on your use case.

If we go forward with returning logits, we should put this feature into the inferencer and make it optional, since it will unnecessarily slow down most other use cases.

malteos:
Sorry for not coming back to you. Other things are keeping me busy these days. 

I close this for now and may reopen it as soon as I continue working on it.

PR #345, opened 2020-05-01
Removing lambda function from AdaptiveModel so the class can be pickable

skirdey:
A class that can be pickled is easier to use in a distributed system where artifacts get serialized and passed around.


tholor:
Makes sense. Thanks for refactoring this!

PR #348, opened 2020-05-05
Update tutorials

bogdankostic:


bogdankostic:
> We should also update the Colab notebooks with the same changes. Let me know if you need permissions here.

Already done! :)

PR #351, opened 2020-05-06
Update PyTorch to 1.5

tanaysoni:


tholor:
Did you verify that the GPUs can be found in our new docker based on `nvidia/cuda:10.2-base-ubuntu18.04`? I know that with the old base image it was not working due to cuda upgrade in pytorch 1.5.0

tanaysoni:
Yes, this image works on V100 GPU.

PR #356, opened 2020-05-08
allows data_silo.calculate_class_weights to support multilabel

renaud:
I was having an issue with using `TextClassificationProcessor(multilabel=True)` together with `data_silo.calculate_class_weights`

I belive this commit solves the issue, please let me know if it's the way to go.

    processor = TextClassificationProcessor(
                                            multilabel=True,
                                            )

    data_silo = DataSilo(
        processor=processor,
        batch_size=batch_size)

    language_model = LanguageModel.load(lang_model)
    prediction_head = MultiLabelTextClassificationHead(
        class_weights=data_silo.calculate_class_weights(task_name="text_classification"),
        num_labels=len(label_list))

### DIFF
--- a/farm/data_handler/data_silo.py
+++ b/farm/data_handler/data_silo.py
@@ -477,7 +477,8 @@ class DataSilo:
             raise Exception("source argument expects one of [\"train\", \"all\"]")
         for dataset in datasets:
             if dataset is not None:
-                observed_labels += [label_list[x[tensor_idx].item()] for x in dataset]
+                for x in dataset:
+                    observed_labels += [label_list[label_id] for label_id in (x[tensor_idx] == 1).nonzero()]
         #TODO scale e.g. via logarithm to avoid crazy spikes for rare classes
         class_weights = list(compute_class_weight("balanced", np.asarray(label_list), observed_labels))
         return class_weights


farm/data_handler/data_silo.py
```diff
@@ -477,7 +477,8 @@ def calculate_class_weights(self, task_name, source="train"):
             raise Exception("source argument expects one of [\"train\", \"all\"]")
         for dataset in datasets:
             if dataset is not None:
-                observed_labels += [label_list[x[tensor_idx].item()] for x in dataset]
+                for x in dataset:
+                    observed_labels += [label_list[label_id] for label_id in (x[tensor_idx] == 1).nonzero()]
```
Timoeller, line 481: That wont work for normal classification since in this case x[tensor_idx] just holds a single scalar in range [0 to number of classes - 1]. So it will evaluate to false in most cases and we will have a much smaller observed_labels array.

What about you get the multilabel flag from the processor and switch between label extraction?

renaud: sounds good. thanks for the prompt feedback

Timoeller:
Changes + fixes merged with #389

Issue #362, opened 2020-05-13
Fix activation function for pooled ELECTRA output


stefan-it:
Hi,

inspired by https://github.com/huggingface/transformers/pull/4257 I just had a look at the original ELECTRA implementation, and it seems that they're using `gelu` instead of `tanh` for getting a pooled output:

https://github.com/google-research/electra/blob/79111328070e491b287c307906701ebc61091eb2/model/modeling.py#L247

and:

https://github.com/google-research/electra/blob/79111328070e491b287c307906701ebc61091eb2/model/modeling.py#L45

Unfortunately, I used `tanh` in the ELECTRA model integration:

https://github.com/deepset-ai/FARM/blob/37591d198813c999afe3766be2fa281af26a25e5/farm/modeling/language_model.py#L1240

I will open a PR to fix that :)

brandenchan:
Implemented in #364

PR #364, opened 2020-05-13
ELECTRA: use gelu for pooled output of ELECTRA model

stefan-it:
Hi,

this PR fixes the activation function for pooled output from the ELECTRA model, to match the original implementation.

`gelu` is now used, wheras e.g. BERT uses `tanh` as activation function. See discussion in #362.


stefan-it:
@brandenchan Would be interesting so see the performance diff :)

brandenchan:
Great! I tested this on an Electra checkpoint and the performance is still good (in fact gelu got 0.2% averaged over 3 germeval tasks than tanh). Our CI is currently crashing but when I tested the branch locally, all tests passed!

PR #382, opened 2020-05-29
Docstring Format fix

PhilipMay:
The docstring formatting has some small bugs. See here:

1. https://farm.deepset.ai/api/running.html#farm.train.EarlyStopping.__init__
2. https://farm.deepset.ai/api/running.html#farm.train.EarlyStopping.check_stopping
3. https://farm.deepset.ai/api/modeling.html#farm.modeling.adaptive_model.AdaptiveModel.forward_lm

The reason is that we need a blank line before the parameter section and the text section. This is fixed here.

### DIFF
--- a/farm/modeling/adaptive_model.py
+++ b/farm/modeling/adaptive_model.py
@@ -426,9 +426,11 @@ class AdaptiveModel(nn.Module, BaseAdaptiveModel):
     def forward_lm(self, **kwargs):
         """
         Forward pass for the language model
+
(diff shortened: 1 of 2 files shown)


PhilipMay:
Also fixed error like this:
```
Warning, treated as error:
/home/mike/dev/git/FARM/farm/train.py:docstring of farm.train.EarlyStopping.check_stopping:6:Field list ends without a blank line; unexpected unindent.
```

Issue #379, opened 2020-05-29
Add possibility to specify text col name in `TextClassificationProcessor`


PhilipMay:
With `label_column_name` it is possible to set the col name of the labels in `TextClassificationProcessor`. IMO it should also be possible to specify the text col name. At the moment it seems to be fixed to `text`.

If you agree and want to I can provide a PR.

Timoeller:
We would love a PR on this one! It is on our roadmap but we didn't find time for it yet.

PhilipMay:
@Timoeller one question that came up while coding this PR:
Under which condition can `self.tasks.values()` contain more than one value? See here:
- https://github.com/deepset-ai/FARM/blob/master/farm/data_handler/processor.py#L463
- https://github.com/deepset-ai/FARM/blob/master/farm/data_handler/processor.py#L508
- and so on...

IMO the `add_task` function only adds exactly one task per `Processor` class.

PhilipMay:
Implemented in #387. Closing this.

PR #387, opened 2020-05-30
Option to specify text col name in `TextClassificationProcessor` and `RegressionProcessor`

PhilipMay:
PR for #379 

## TODO
- [x] check if and how tests can be written -> added parametrize to the tests


farm/data_handler/utils.py
```diff
@@ -64,11 +64,12 @@ def read_tsv(filename, rename_columns, quotechar='"', delimiter="\t", skiprows=N
     # let's rename our target columns to the default names FARM expects:
     # "text": contains the text
     # "text_classification_label": contains a label for text classification
-    columns = ["text"] + list(rename_columns.keys())
+    columns = list(rename_columns.keys())
```
Timoeller, line 67: The code that was already present in FARM to rename columns of a pandas DF is a shame : ) and should be changed.

While we are at it: could you use the pandas functionality for renaming the columns here, please? 
`df.rename(columns=rename_columns, inplace=True)`

Timoeller:
Could you do a dummy commit to restart the CI, please? (I do not have access to the pipeline, today is public holiday in Germany...)

PhilipMay:
rebased master

PhilipMay:
Fixed `test_processor_saving_loading` the same way as https://github.com/deepset-ai/FARM/pull/390 does.

PhilipMay:
The CI system was gracious to me and all tests are green now. :-)

Can you please do a review?

Thanks
Philip

farm/data_handler/utils.py
PhilipMay: I made the changes.

farm/data_handler/processor.py
```diff
@@ -454,13 +465,17 @@ def __init__(
...
         else:
             logger.info("Initialized processor without tasks. Supply `metric` and `label_list` to the constructor for "
                         "using the default task or add a custom task later via processor.add_task()")
 
     def file_to_dicts(self, file: str) -> [dict]:
-        column_mapping = {task["label_column_name"]: task["label_name"] for task in self.tasks.values()}
+        column_mapping = {}
+        for task in self.tasks.values():
+            column_mapping[task["label_column_name"]] = task["label_name"]
+            column_mapping[task["text_column_name"]] = "text"
```
Timoeller, line 478: This might lead to unforeseen problems if a column "text" already exists.
I know we also do not check for the existence of task["label_name"] but now could be a good time to improve.

What about we check if any of task["label_name"] or "text" is already inside the dataframe and log a warning?

PhilipMay: Good point. I will come up with a solution...

PhilipMay: Ok - so `column_mapping` is a dict which maps "col_name_in_tsv_file" to "col_name_farm_wants". 
In `read_tsv` (`utils.py`) **all** of the tsv file is loaded with pandas. Then this is done:

```python
    columns = list(rename_columns.keys())
    df = df[columns]
```

This takes the keys from the dict as a list. The keys are the "col_name_in_tsv_file". So after that I have a pandas frame with only those cols I want to have. Then they are renamed: 
`df.rename(columns=rename_columns, inplace=True)`.

It the dict maps from `text` to `text` that will be no problem. It just selects the text col and then renames `text` to `text`. This renaming is nonsence but works. It could be removed but creates more code.

Before calling `df.rename(columns=rename_columns, inplace=True)` I could remove all entries from the dict where the key is the same as the value. Should I do that? What do you think? IMO it is not necessary.

But while discussing this it brought me to an other idea:

In `pd.read_csv` I could only read only those cols I need. Those in `list(rename_columns.keys())`. That would have the advantage that the code crashes early when the user mistyped a col name that is not in the data.

PhilipMay:
Made this change discussed above.

I am passing `usecols=columns_needed` to `pd.read_csv`. to read only those cols we need. Right from the beginning. The line `df = df[columns_needed]` is not needed anymore.

The advantage is memory usage and early crashing. If one of the cols does not exist it crashes with a meaningful error. Example: `ValueError: Usecols do not match columns, columns expected but not found: ['text_other_']`

PhilipMay:
rebased master (mainly to trigger CI again)

PhilipMay:
YAY! CI is green. :-)

farm/data_handler/processor.py
Timoeller: Thanks for this detailed analysis. Your proposed changes make sense and improve upon functionality.

Timoeller:
CSV data reading is handled much more user friendly now.
I will also adjust the read_tsv_sentence_pair() to reflect these changes.

Thanks @PhilipMay for your effort.

Issue #388, opened 2020-05-31
Problems with `**kwargs` in `TextClassificationProcessor`


PhilipMay:
The `TextClassificationProcessor` has an `**kwargs` param in its constructor. See here: https://github.com/deepset-ai/FARM/blob/258a76fface7652248530bdd9c841d6022c89af5/farm/data_handler/processor.py#L386

The `**kwargs` param is not used anywhere and should be removed. The reason is that can "eat" wrong parameter names without notice. Exactly this happens here:
https://github.com/deepset-ai/FARM/blob/master/test/test_processor_saving_loading.py#L18

Where `metrics` is wrong. It should be `metric` without `s`. Also `columns` is an unknown param and just "eaten" by `**kwargs`.

What do you think? As always I could provide a PR.

Timoeller:
Agreed, it is nasty when this happens unexpectedly and we should avoid generic kwargs wherever possible.

I was going through the code to check if it is feasible and realized we get into trouble when generically loading. I might have a working solution in #390 but will test it more thoughoughly later on.

Thanks for raising this issue!

PR #390, opened 2020-05-31
WIP: Remove generic **kwargs parameters from Processor subclasses

Timoeller:
Addresses #388 

We need to make sure only fitting params are passed when loading a processor.


farm/data_handler/processor.py
```diff
@@ -383,7 +384,6 @@ def __init__(
         header=0,
         proxies=None,
         max_samples=None,
-        **kwargs
```
PhilipMay, line 386: This change is right IMO. But it breaks this test: https://github.com/deepset-ai/FARM/blob/5e74f8ad881a08c35bd23ee7561ed20271b2cf0c/test/test_processor_saving_loading.py#L18

The test should be fixed.

Timoeller: Good point, I adjusted the test.

I am happy about the kwargs removal. This test is pretty outdated and should not have been working.

Timoeller:
Loading isn't working in all cases with the proposed changes.

The load functionality of the processor has hardcoded args like data_dir etc. These hardcoded parameter assignments need to be removed.

Issue #385, opened 2020-05-30
CI Pipeline seems to be broken


PhilipMay:
For some reason the CI errors out. See here:

https://dev.azure.com/deepset/FARM/_build/results?buildId=285&view=logs&j=5f32d555-91de-5b47-aa96-4f80d9326a6b&t=813bfe30-cb15-55d3-e33c-c7c8b5f23c95&l=81

This is from the last commit: https://github.com/deepset-ai/FARM/commit/258a76fface7652248530bdd9c841d6022c89af5

In the PR (before merging) the CI was all green: https://github.com/deepset-ai/FARM/pull/382

What is the reason here?

Timoeller:
So the error: `Bash exited with code '137'` has to do with memory issues. We recently changed from Travis CI to Azure pipelines because we thought this fixes OOM issues. @tanaysoni could you please have a look at it?

PhilipMay:
We should reduce the memory usage of the `test_question_answering.py ` test. Maybe use a smaller model then `distilbert-base-uncased`?

PhilipMay:
Fixed by: https://github.com/deepset-ai/FARM/commit/32fe61ba17870e0f43b42b3e3d179dfa8fb4d910

Closing this.

PhilipMay:
Well, there still seems to be a problem which does not always happen but just from time to time. See here: https://dev.azure.com/deepset/FARM/_build/results?buildId=303&view=logs&j=5f32d555-91de-5b47-aa96-4f80d9326a6b&t=813bfe30-cb15-55d3-e33c-c7c8b5f23c95&l=87

This happened after I rebased master and got the Update to pytorch 1.5.0.
Could you please investigate this?

I will reopen this. Issue.

Thanks
Philip

PhilipMay:
More strange things happening:
- `test_ner.py` crash: https://dev.azure.com/deepset/FARM/_build/results?buildId=311&view=logs&j=5f32d555-91de-5b47-aa96-4f80d9326a6b&t=813bfe30-cb15-55d3-e33c-c7c8b5f23c95&l=81
- complete f*up: https://github.com/deepset-ai/FARM/runs/728689459

Timoeller:
:D jap, confirming there is no such thing as a free lunch and our azure CI pipeline is on a free plan for now...
I spoke to our engineer and the good news is the CI issues - especially the OOM issue - popped up a while ago and might not be related to our PRs. Fixing these issues properly will require some more time on our end.

[...]

PR #391, opened 2020-05-31
CI OOM Fix with smaller Model

PhilipMay:
CI is running OOM. See here #385 

This might be a fix. It is using a smaller model (`distilbert-base-cased` and `albert-base-v2`). See here: https://huggingface.co/transformers/pretrained_models.html

### DIFF
--- a/test/test_lm_finetuning.py
+++ b/test/test_lm_finetuning.py
@@ -23,7 +23,7 @@ def test_lm_finetuning(caplog):
     n_epochs = 1
     batch_size = 1
     evaluate_every = 2
-    lang_model = "bert-base-cased"
+    lang_model = "distilbert-base-cased"
 
     tokenizer = Tokenizer.load(
         pretrained_model_name_or_path=lang_model, do_lower_case=False
--- a/test/test_question_answering.py
+++ b/test/test_question_answering.py
@@ -22,7 +22,7 @@ def test_qa(caplog=None):
     batch_size = 2
     n_epochs = 1
     evaluate_every = 4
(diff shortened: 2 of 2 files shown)


PhilipMay:
Well - it can not load `distilbert-base-cased`. Is that "normal"?

Timoeller:
FARM can load "distilbert-base-uncased" but not for LM finetuning.

[In the corresponding LM finetuning test](https://github.com/deepset-ai/FARM/blob/master/test/test_lm_finetuning.py#L43) we are loading:
- `language_model = LanguageModel.load(lang_model)` works with distilbert
 - `lm_prediction_head = BertLMHead.load(lang_model)` doesnt work since we load a BertConfig instead of a disttilbertconfig...

In any case, we already resolved the OOM CI issues when moving to azure pipelines. I suspect something changed there under the hood and we need to take a closer look. We might be able to resolve this by tomorrow.

PhilipMay:
Could you temporarily deactivate this test? 

ˋ@pytest.mark.skip(reason="no way of currently testing this")ˋ

Timoeller:
I am working on a fix in #392. When the test pass there we can rerun the pipelines for the other PRs

Timoeller:
Hey @PhilipMay updating to torch==1.5.0 seems to resolve OOM issues. Lets rerun the tests and then merge the other open PRs.

I would prefer to close this PR since 
- we want to test lm finetuning with bert directly instead of distilbert
- question answering with distilbert seems fine to me (and from our tests albert does not really reduce memory consumption)

PhilipMay:
Ok. 👍

Issue #352, opened 2020-05-06
RuntimeError: Expected object of scalar type Float but got scalar type Double for argument #3 'weight' in call to _thnn_nll_loss_forward


renaud:
**Describe the bug**
i'm performing classification, using the script below. 
when installing "with git" (dbcf91d9a35ac54fb31d890d90ce0e28dd93cda6), I get the error below.
I then installed the 0.4.3 version throught pip and it worked.
Can you reproduce or is it just me?

    set_all_seeds(seed=42)
    n_epochs   = 8
    batch_size = 16
    acc_steps  = 2
    evaluate_every = 128
    lang_model = "bert-base-cased"
    do_lower_case = False
    learning_rate = 2e-5
    max_seq_len = 64
    use_amp = None

    device, n_gpu = initialize_device_settings(use_cuda=True, use_amp=use_amp)

    tokenizer = Tokenizer.load(pretrained_model_name_or_path=lang_model, do_lower_case=do_lower_case)

    label_list = ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17"]
    metric = "f1_macro"
    processor = TextClassificationProcessor(tokenizer=tokenizer,
                                            max_seq_len=max_seq_len,
                                            data_dir=Path(DATA),
                                            dev_filename="dev.tsv",
                                            label_list=label_list,
                                            metric=metric,
                                            label_column_name="sdg_label"
                                            )

    data_silo = DataSilo(
        processor=processor,
        batch_size=batch_size)

    language_model = LanguageModel.load(lang_model)
    prediction_head = TextClassificationHead(
        class_weights=data_silo.calculate_class_weights(task_name="text_classification"),
        num_labels=len(label_list))

    model = AdaptiveModel(
        language_model=language_model,
        prediction_heads=[prediction_head],
        embeds_dropout_prob=0.1,
        lm_output_types=["per_sequence"],
        device=device)

    # 5. Create an optimizer
    model, optimizer, lr_schedule = initialize_optimizer(
        model=model,
        learning_rate=learning_rate,
        device=device,
        n_batches=len(data_silo.loaders["train"]),
        n_epochs=n_epochs,
        use_amp=use_amp,
        grad_acc_steps=acc_steps)
 

**Error message**

    Train epoch 0/8 (Cur. train loss: 0.0000):   0%|          | 0/2370 [00:00<?, ?it/s]

    ---------------------------------------------------------------------------

[...]

PR #392, opened 2020-06-01
Update pytorch to 1.5.0

Timoeller:
Testing if updating to newer pytorch resolves our CI OOM issues.

Fixing pytorch==1.4.0 in 69d90a0f7121679d2dbaf20cad83b73d8af2911b introduced [some unknown CI problems](https://dev.azure.com/deepset/FARM/_build/results?buildId=281&view=results)


Timoeller:
Ok, this seems to fix the CI errors. Will need to investigate on azure side what caused it.

Will now test the class weights issue #352 and this newer pytorch version.

Timoeller:
https://github.com/pytorch/pytorch/pull/30486 introduced correct casting of numpy dtypes. So we need to cast our class weights to an appropriate type.

Issue #366, opened 2020-05-13
pip install on Windows


ftesser:
**Describe the bug**
Installing FARM on Windows by pip install does not work because pip install torch does not work on Windows. 

**Error message**
```bash
pip install farm
Collecting farm
  Using cached farm-0.4.3.tar.gz (153 kB)
Requirement already satisfied: setuptools in c:\programs\python\python37\lib\site-packages (from farm) (46.2.0)
Requirement already satisfied: wheel in c:\programs\python\python37\lib\site-packages (from farm) (0.34.2)
ERROR: Could not find a version that satisfies the requirement torch==1.4.0 (from farm) (from versions: 0.1.2, 0.1.2.post1, 0.1.2.post2)
ERROR: No matching distribution found for torch==1.4.0 (from farm)
```

**Expected behavior**
Install torch package also on Windows.

**Additional context**
I know that this is not really a bug of FARM, but I have some suggestions to allow the dependency installation works also on Windows.

The suggestion cames from this discussion:
https://discuss.pytorch.org/t/how-to-add-pytorch-as-a-requirement-for-a-windows-python-package/67353/5

Adding some windows specific code on setup.py can allow the auto-installalation of the dependecies when setup.py is used.

Moreover I like to have both solutions working:
 - auto install dependencies with setup.py 
 - pip install -r requirements.txt 

For this second option my proposal is that setup.py can produce (just in teh windows case) a modifed version of requirements.txt adding -f https://download.pytorch.org/whl/cpu/torch_stable.html on the torch line.

I agree with the author that I am not happy to do this (the best should be that torch fix that, I do not undertsand why torch for Windows can not be availabe on pypi.org) but at least this give the possibility of automatize installations on different OS in some way.

If you agree on this solution I can prepare a merge request.

**To Reproduce**
```bash
pip install farm
```
**System:**
 - OS: Windows
 - GPU/CPU: CPU
 - FARM version:

Timoeller:
Hey @ftesser sorry for the late reply.

I think having a clean solution inside FARM would be great. I think the most important point is that installation on other systems do not break. If you already have changes please create a WIP merge request so we can get this functionality into farm.

ftesser:
Hi @Timoeller, sorry for the late reply as well. I have just submitted a PR.

tanaysoni:
Fixed by #393.

PR #393, opened 2020-06-01
Managing pytorch pip find-links directive

ftesser:
This is a solution for #366.

This solution includes:
 * the addition of  --find_links in the requirements.txt file (this is the suggested method also for Linux as suggested in https://pytorch.org/get-started/locally/)
 * the parsing of requirements.txt in setup.py to manage this new directive (https://stackoverflow.com/questions/57689387/equivalent-for-find-links-in-setup-py)

Note: I have not included in the new parsing function (parse_requirements) of the requirements.txt  the condition:
```python
if ((x.strip()[0] != "#") and (len(x.strip()) > 3) and "-e git://" not in x)
```
because I did not see its actual use . But if it is necessary we can add inside the parse_requirements function.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@
 setuptools
 wheel
 # PyTorch
+--find-links=https://download.pytorch.org/whl/torch_stable.html
 torch==1.5.0
 # progress bars in model download and training scripts
 tqdm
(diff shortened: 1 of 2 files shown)


tanaysoni:
Hi @ftesser, thank you for the PR. It looks good-to-go.

PR #396, opened 2020-06-02
Add support for Camembert-like models

### LINKED ISSUE #386: Support for CamemBERT-like model types
It would be nice to be able to use CamemBERT like models in the experiment environment. UmBERTo, the italien model is also like that.

Please add it to the experiment framework :)

**Additional context**
Information about CamemBERT: https://huggingface.co/transformers/model_doc/camembert.html

bogdankostic:
Resolves #386.


farm/modeling/adaptive_model.py
```diff
@@ -256,7 +256,6 @@ def __init__(
         self.lm_output_types = (
             [lm_output_types] if isinstance(lm_output_types, str) else lm_output_types
         )
-        assert len(self.lm_output_types) == len(self.prediction_heads)
```
tholor, line 259: Why are we getting rid of this assertion here?

Timoeller: It fails during embedding extraction like described in #397

PR #400, opened 2020-06-04
Add stalebot

tholor:
Adding stalebot to automatically close old, inactive issues

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
tanaysoni, line 8: Do we want to exempt the `enhancement` label?

tholor: No, I think after 60 days without any activity we usually don't need them. There's quite many ideas for enhancements that would be nice in theory but are out of scope. There's also still the option to prevent them  being closed by posting something within 14 days after the stale bot has marked them as stale.

PR #405, opened 2020-06-10
Implement Prediction Objects for Question Answering

brandenchan:


farm/modeling/prediction_head.py
```diff
@@ -1395,14 +1428,14 @@ def reduce_preds(self, preds):
...
+                seen[(qa_answer.offset_answer_start, qa_answer.offset_answer_end)] = qa_answer
             else:
-                seen_score = seen[(span.start, span.end)].score
-                if span.score > seen_score:
-                    seen[(span.start, span.end)] = span
-        return [span for span in seen.values()]
+                seen_score = seen[(qa_answer.offset_answer_start, qa_answer.offset_answer_end)].score
+                if qa_answer.score > seen_score:
+                    seen[(qa_answer.offset_answer_start, qa_answer.offset_answer_end)] = qa_answer
+        return [qa_answer for qa_answer in seen.values()]
```
Timoeller, line 1438:
```suggestion
        return list(seen.values())
```

farm/modeling/predictions.py
```diff
@@ -1,77 +1,166 @@
...
+        else:
+            self.answer = string
+            assert self.offset_answer_end >= 0
+            assert self.offset_answer_start >= 0
+
+    def to_list(self):
+        return [self.answer, self.offset_answer_start, self.offset_answer_end, self.score, self.sample_idx]
+
+
+class QAAnswerOLD:
```
Timoeller, line 57: Can we remove this?

farm/modeling/predictions.py
```diff
@@ -1,77 +1,166 @@
...
+                 prediction,
+                 context):
+        self.id = id
+        self.prediction = prediction
+        self.context = context
+
+    def to_json(self):
+        raise NotImplementedError
+
+class QAAnswer(BaseModel):
```
Timoeller, line 18: 1. Suggestion: should we rename this to QACandidate, to better differentiate it from QAPred?

2. Why doesnt this class have an init method and no attributes?

farm/modeling/predictions.py
```diff
@@ -1,77 +1,166 @@
...
+            assert self.offset_answer_start >= 0
 
-    def __repr__(self):
-        return str(self)
 
-class DocumentPred:
-    """ Contains a collection of Span predictions for one document. Used in Question Answering. Also contains all
-    attributes needed to generate the appropriate output json"""
+
+class QAPred(Pred):
```
Timoeller, line 129: Can we add type annotations here, too please?

Timoeller:
Please also adjust test_inference.py  since these test arent running through.

farm/modeling/predictions.py
brandenchan: yup

farm/modeling/predictions.py
brandenchan: QACandidate would be better. No init methods is because it is a pydantic.BaseModel object. That's just the syntax they use

farm/modeling/predictions.py
brandenchan: Yes sorry forgot to do this object too

PR #406, opened 2020-06-11
Quick fix CI problems with OOM and unclosed worker pool

PhilipMay:
Temporary fix for CI problems with OOM and unclosed worker pool. Also see: #385 and #403 

This is just a quick fix and should be replaced by a real fix. See discussion in #403


tanaysoni:
Hi @PhilipMay, thank you for the PR!

PR #424, opened 2020-06-25
remove hardcoded seeds from trainer

### LINKED ISSUE #423: Trainer fixes all seeds to 39
**Describe the bug**
In Trainer.train() seeds are hardcoded to 39 by calling set_all_seeds(seed=39) (two times).
Thus any calls to set_all_seeds in training scripts are ignored.

**Error message**
None

**Expected behavior**
Calling set_all_seeds from arbitrary training scripts works as intended and is not ignored. 

**To Reproduce**
start arbitrary training

**System:**
 - OS: all
 - GPU/CPU: all
 - FARM version: since 5151b36bbdc77c9a571a57e5c571b8c45f87acf5

tstadel:
resolves #423 by removing calls to set_all_seeds in Trainer.train()

### DIFF
--- a/farm/train.py
+++ b/farm/train.py
@@ -249,7 +249,6 @@ class Trainer:
         evalnr = 0
         loss = 0
         resume_from_step = self.from_step
-        set_all_seeds(seed=39)
 
         if self.local_rank in [0, -1]:
             logger.info(f"\n {GROWING_TREE}")
@@ -274,7 +273,6 @@ class Trainer:
                     else:
                         continue
 
-                set_all_seeds(seed=39)
                 progress_bar.set_description(f"Train epoch {epoch}/{self.epochs-1} (Cur. train loss: {loss:.4f})")
 
                 # Only for distributed training: we need to ensure that all ranks still have a batch left for training


tholor:
Thanks for flagging this and proposing a fix @tstadel ! 

The reason why we set the seeds there was to gain full reproducibility of training with checkpointing vs. without. 
If you load a checkpoint, we will "fast-forward" the DataLoader to the batch where you stopped in your last run (e.g step 100). During this fast-forwarding, the random state will change as there are operations that modify it (e.g. finding a random next sentence). So even though we load the random states from the checkpoint, they will be altered after the fast-forwarding.

I agree that they shouldn't be hardcoded to specific values (=39), but do you see a particular down-side of having them there? We could, for example, add an arg "seed" to the Trainer.__init__() and use this value instead.

tstadel:
Thanks for the explanation!

Nevertheless I have to say, it just looks pretty odd to me to set the seed in every step... Even if we introduce an arg to the Trainer, calls to `set_all_seeds()` from upstream code would be effectively ignored. So in order to reproduce results from such code without having to change it, the seed setting within `train()` should be optional.

Is it just simply cumbersome to restore the seeds stored in the checkpoint after fast-forwarding or is there another reason we can't do that?

tholor:
> Is it simply too cumbersome to restore the seeds stored in the checkpoint after fast-forwarding or is there another reason we can't do that?

It's not really complex, but it would probably make the code a bit less readable as we don't have access to the checkpoint anymore after the initial loading via the Trainer's class method. One option would be to pass the checkpoint path as an attribute  to the Trainer instance and then loading the random states like here: 

https://github.com/deepset-ai/FARM/blob/0098819e4fb4df4c9fba060a8272ebf3bf9e58d2/farm/train.py#L477-L482

Thinking more about this, I agree that the current situation is not ideal as recent research also shows quite some variance of performance for different seeds in downstream tasks. This "seed hacking" is not possible right now.  

Thus, my suggestion is to:
- short-term: get rid of these seeds
- mid/long-term: load the random states from the checkpoint after fast-forwarding

Let me know if you are willing to work on the midterm solution.

tstadel:
Yes, I'd like to work on the midterm solution. I'm trying to find some time for this till the end of the week.

tholor:
Perfect, thanks!

PR #429, opened 2020-06-26
Fix and document Inferencer usage and pool handling

PhilipMay:
This is the 2nd part of PR #403 

## TODO
- [x] check examples 
- [x] check call in `fit_s3e_on_corpus`
- [x] check other documentation
- [x] sould other code be changed to close pool
- [x] must be merged with PR #430
- [ ] update `docs/basic_usage.rst`
- [ ] update `docs/examples.rst`
- [ ] update `tutorials/1_farm_building_blocks.ipynb`
- [ ] write explicit example in doc and tell the user how to use this

### DIFF
--- a/examples/conversion_huggingface_models.py
+++ b/examples/conversion_huggingface_models.py
@@ -17,7 +17,13 @@ def convert_from_transformers():
 
     # CASE 2: INFERENCER
     # Load Inferencer from transformers, incl. model & tokenizer (-> just get predictions)
-    nlp = Inferencer.load("deepset/bert-large-uncased-whole-word-masking-squad2", task_type="question_answering")
+    # Warning! If you use multiprocessing and open a pool by passing
+    # `None` or an integer greater zero to `num_processes` please make
+    # sure to close the pool again by calling `close_multiprocessing_pool`.
+    # The garbage collector will not do this for you!
+    nlp = Inferencer.load("deepset/bert-large-uncased-whole-word-masking-squad2",
+                          task_type="question_answering",
+                          num_processes=0)
 
     # run predictions
     QA_input = [{"questions": ["Why is model conversion important?"],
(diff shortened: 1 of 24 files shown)


tanaysoni:
Hi @PhilipMay, I added a suggestion commit with a call to the `close_multiprocessing_pool()` each time we use the Inferencer. In that case, we might not need the warning comments. What do you think?

PhilipMay:
> Hi @PhilipMay, I added a suggestion commit with a call to the `close_multiprocessing_pool()` each time we use the Inferencer. In that case, we might not need the warning comments. What do you think?

Well, that does not fully work because you would have to wrap the code into a try and finally block to even close the pool when an exception is raised.

PhilipMay:
And to avoid this kind of "ugly and complicated" code I would better not open a pool... So the example is easy to read.

tanaysoni:
Hi @PhilipMay, 

In the case of exception in the Inferencer, the example scripts would crash, terminating all the Python subprocesses. For instance, the following script should exit clean without any zombie processes.

```python
results = inferencer.inference_from_dicts(qa_input)
print(results)
raise Exception  # scripts exits here
inferencer.close_multiprocessing_pool()
```

PhilipMay:
@tanaysoni Yes I agree. In the example it does not cause any direct problems. But IMO an example should be something that can be taken and put into something bigger. If placed in a real life server environment this might cause an ugly memory leak if not handled correctly. But we already had this discussion on an other PR if I remember correctly.

PhilipMay:
So what do you suggest we can do with this uncompleted PR? In which direction should we go?

tanaysoni:
Hi @PhilipMay, IMO we add a call to `close_multiprocessing_pool()` in each example along with a single line comment mentioning it's important to close the pool. Additionally, we can write a detailed example explaining more details about the Pool behavior. What do you think?

PhilipMay:
Ok. Thanks. 👍

Issue #428, opened 2020-06-26
Loading of saved model results in Unexpected key(s) in state_dict: "loss_fct.weight"


Timoeller:
**Describe the bug**
We cannot load a Prediction Head with class weights.

**Error message**
RuntimeError: Error(s) in loading state_dict for TextClassificationHead:
	Unexpected key(s) in state_dict: "loss_fct.weight". 

I suspect this bug was introduced by updating to pytorch version 1.5.0.
Unfortunately we do not have tests covering this use case, so we cannot pin the cause exactly.

PhilipMay:
See #422 for a detailed exception.

PhilipMay:
See RP #431 for fix.

Issue #422, opened 2020-06-24
Bug in example `doc_classification_with_earlystopping.py`.


PhilipMay:
There is a bug executing `doc_classification_with_earlystopping.py` on colab. See full notebook here: https://gist.github.com/PhilipMay/8b042f713603e68deb5519fb7776d128

```
[...]
<ipython-input-3-66a7ddd46d8d> in doc_classification_with_earlystopping()
    128 
    129     # 7. Let it grow
--> 130     trainer.train()
    131 
    132     # 8. Hooray! You have a model.

/content/FARM/farm/train.py in train(self)
    352             logger.info("Restoring best model so far from {}".format(self.early_stopping.save_dir))
    353             lm_name = self.model.language_model.name
--> 354             model = AdaptiveModel.load(self.early_stopping.save_dir, self.device, lm_name=lm_name)
    355             model.connect_heads_with_processor(self.data_silo.processor.tasks, require_labels=True)
    356 

/content/FARM/farm/modeling/adaptive_model.py in load(cls, load_dir, device, strict, lm_name, processor)
    321         ph_output_type = []
    322         for config_file in ph_config_files:
--> 323             head = PredictionHead.load(config_file, strict=strict)
    324             prediction_heads.append(head)
    325             ph_output_type.append(head.ph_output_type)

/content/FARM/farm/modeling/prediction_head.py in load(cls, config_file, strict, load_weights)
    116             model_file = cls._get_model_file(config_file=config_file)
    117             logger.info("Loading prediction head from {}".format(model_file))
--> 118             prediction_head.load_state_dict(torch.load(model_file, map_location=torch.device("cpu")), strict=strict)
    119         return prediction_head
    120 

/usr/local/lib/python3.6/dist-packages/torch/nn/modules/module.py in load_state_dict(self, state_dict, strict)
    845         if len(error_msgs) > 0:
    846             raise RuntimeError('Error(s) in loading state_dict for {}:\n\t{}'.format(
--> 847                                self.__class__.__name__, "\n\t".join(error_msgs)))
    848         return _IncompatibleKeys(missing_keys, unexpected_keys)
    849 

RuntimeError: Error(s) in loading state_dict for TextClassificationHead:
	Unexpected key(s) in state_dict: "loss_fct.weight". 
```

If you could provide a workaround to avoid the exeption I would be very happy. As always I am willing to start a PR but will need some help with debugging.

Timoeller:
[...]

PR #431, opened 2020-06-27
Fix loading of saved models with class weights

PhilipMay:
This is a PR regarding #428 and #422 

## TODO
- [x] implement test
- [x] fix bug
- [x] review potential sideeffects

### DIFF
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -280,6 +280,11 @@ class TextClassificationHead(PredictionHead):
         self.ph_output_type = "per_sequence"
         self.model_type = "text_classification"
         self.task_name = task_name #used for connecting with the right output of the processor
+
+        if type(class_weights) is np.ndarray and class_weights.ndim != 1:
+            raise ValueError("When you pass `class_weights` as `np.ndarray` it must have 1 dimension! "
+                             "You provided {} dimensions.".format(class_weights.ndim))
+
         self.class_weights = class_weights
 
         if class_weights is not None:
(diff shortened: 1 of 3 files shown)


PhilipMay:
An easy fix would be to pass `strict=False` at this location:
https://github.com/deepset-ai/FARM/blob/451bc0bdd6146f5cbd8b2ca5047814874de717de/farm/modeling/adaptive_model.py#L323

This would avoid the exception but I am not sure if it fixed the underlying problem.

PhilipMay:
The bug also happens when PyTorch 1.4.0 is used. I tested it. This is very strange. What did FARM do to introduce this bug? I am pretty sure that the early stopping example was ok 3 weeks ago...

PhilipMay:
Weight is not correctly set at loading. See screenshot.

Screenshot from 2020-06-27 12-39-44

PhilipMay:
Well, long debugging, quick fix. :-)
IMO it can be reviewed and merged when ok.

PhilipMay:
Because we convert the class_weights to a list I added a check that the numpy array has exactly one dimension.

PhilipMay:
> Calling json.dump on numpy class weights should result in TypeError: Object of type 'ndarray' is not JSON serializable
> Weird that this error happens silently and class weights are simply omitted when saving the Prediction Head config. I might look into this another time.

It is because you checked it if it can be coverted to json:
https://github.com/deepset-ai/FARM/blob/master/farm/modeling/prediction_head.py#L91

Issue #441, opened 2020-07-02
How do I do text classification with multiple heads?


PhilipMay:
Hi,
I have a dataset with two columns of label.
- one column encodes the first class (always one of 4)
- one column encodes the 2nd class (always one of 5)

So I want to use 2 heads to predict the first and 2nd class.
How do I connect the data (TextClassificationProcessor) with the 2 heads?
I saw an example of multi label classification. But that does not exactly fit my need.
Is that even possible with the current implementation of FARM?

Thanks
Philip

tholor:
Hi, 
While I personally haven't worked on this for a while, the basic functionality should be in place. 
Please have a look at the [release notes of v0.2.2](https://github.com/deepset-ai/FARM/releases/tag/0.2.2), where we included an example of how to add multiple tasks. Hope this helps!

PhilipMay:
> Please have a look at the [release notes of v0.2.2](https://github.com/deepset-ai/FARM/releases/tag/0.2.2), where we included an example of how to add multiple tasks. Hope this helps!

Ahh ok. Thanks. I will check that because this PR from me might not be needed: #442

Timoeller:
Closing this issue now as well.

PR #442, opened 2020-07-02
Enable TextClassificationProcessor to work with multiple prediction heads

PhilipMay:
Add functionality to the TextClassificationProcessor to work with multiple prediction heads.

## TODO
- [ ] add example


PhilipMay:
Do should we just close this PR since it can be handled this way: https://github.com/deepset-ai/FARM/issues/441#issuecomment-653424408

PR #444, opened 2020-07-06
WIP: Remove Asserts

brandenchan:
The idea of this PR generally is to remove assertion statement so that running systems do not get interrupted by errors. What used to be asserts will, where appropriate, be converted into logging.error() or logging.warning(). 

This PR focuses only on QA related components. Asserts may still be present if they can be caught by try catch statements such as Processor._featurize_samples() or Processor._init_samples_in_baskets().

### DIFF
--- a/farm/modeling/adaptive_model.py
+++ b/farm/modeling/adaptive_model.py
@@ -185,13 +185,13 @@ class BaseAdaptiveModel:
         model_files.sort()
         config_files.sort()
 
-        if strict:
+        if strict and (len(model_files) != len(config_files)):
             error_str = (
                 f"There is a mismatch in number of model files ({len(model_files)}) and config files ({len(config_files)})."
                 "This might be because the Language Model Prediction Head "
                 "does not currently support saving and loading"
             )
-            assert len(model_files) == len(config_files), error_str
+            logger.warning(error_str)
         logger.info(f"Found files for loading {len(model_files)} prediction heads")
 
         return model_files, config_files
@@ -343,10 +343,11 @@ class AdaptiveModel(nn.Module, BaseAdaptiveModel):
         all_losses = []
         for head, logits_for_one_head in zip(self.prediction_heads, logits):
             # check if PredictionHead connected to Processor
-            assert hasattr(head, "label_tensor_name"), \
-                (f"Label_tensor_names are missing inside the {head.task_name} Prediction Head. Did you connect the model"
-                " with the processor through either 'model.connect_heads_with_processor(processor.tasks)'"
-                " or by passing the processor to the Adaptive Model?")
+            if not hasattr(head, "label_tensor_name"):
+                error_str = f"Label_tensor_names are missing inside the {head.task_name} Prediction Head. Did you connect the model"\
+                            " with the processor through either 'model.connect_heads_with_processor(processor.tasks)'"\
+                            " or by passing the processor to the Adaptive Model?"
+                logger.error(error_str)
             all_losses.append(head.logits_to_loss(logits=logits_for_one_head, **kwargs))
         return all_losses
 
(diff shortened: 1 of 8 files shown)


farm/modeling/prediction_head.py
```diff
@@ -374,7 +374,13 @@ def formatted_preds(self, logits=None, preds=None, samples=None, return_class_pr
         is needed since at inference, the order of operations is very different depending on whether we are performing
         aggregation or not (compare Inferencer._get_predictions() vs Inferencer._get_predictions_and_aggregate())"""
 
-        assert (logits is not None) or (preds is not None)
+        logit_input = logits is not None
+        preds_input = preds is not None
+
+        if logit_input and preds_input:
```
Timoeller, line 380: syntax seems confusing:
how about 

```
if logits and preds: logger.warning("Both logits and preds have been passed as input to the TextClassificationHead")

if (logits is None) and (preds is None): logger.error("Neither logits nor preds have been passed as input to the TextClassificationHead")
```

farm/modeling/adaptive_model.py
```diff
@@ -343,10 +343,11 @@ def logits_to_loss_per_head(self, logits, **kwargs):
         all_losses = []
         for head, logits_for_one_head in zip(self.prediction_heads, logits):
             # check if PredictionHead connected to Processor
-            assert hasattr(head, "label_tensor_name"), \
-                (f"Label_tensor_names are missing inside the {head.task_name} Prediction Head. Did you connect the model"
-                " with the processor through either 'model.connect_heads_with_processor(processor.tasks)'"
-                " or by passing the processor to the Adaptive Model?")
+            if not hasattr(head, "label_tensor_name"):
```
Timoeller, line 346: Not sure about this one. Can the model work without this? e.g. if it is in inference mode.

If so we should remove the assert, if not and the functionality must break further downstream, then lets keep the assert here

farm/modeling/prediction_head.py
```diff
@@ -1200,12 +1202,19 @@ def formatted_preds(self, logits=None, preds=None, baskets=None, **kwargs):
...
         """
 
+        # Log cases where this method receives the wrong kind of input
+        logit_input = logits is not None
+        preds_input = preds is not None
+
+        if logit_input:
+            logger.warning("QuestionAnsweringHead.formatted_preds() received logit input when it only expects pred input")
+        if not preds_input:
+            logger.warning("QuestionAnsweringHead.formatted_preds() did not receive the preds input it expects")
```
Timoeller, line 1212: this should be at least an logger.error or even kept as assert, since the app breaks without preds

farm/modeling/prediction_head.py
```diff
@@ -1217,8 +1226,6 @@ def formatted_preds(self, logits=None, preds=None, baskets=None, **kwargs):
         # are prediction spans
         preds_d = self.aggregate_preds(preds, passage_start_t, ids, seq_2_start_t)
 
-        assert len(preds_d) == len(baskets)
```
Timoeller, line 1220: we should through a logger error here, since we need as many preds_d (on document level) as we have baskets.
If that is not the case it isnt necessarily breaking a cosuming app. Example: haystack retrieves 10 docs, one of those documents is malformatted so it cannot contain preds. Still we want to give back preds_d on all other 9 docs so the reader produces some answer

farm/modeling/tokenization.py
```diff
@@ -316,7 +316,6 @@ def _words_to_tokens(words, word_offsets, tokenizer):
             else:
                 start_of_word.append(False)
 
-    assert len(tokens) == len(token_offsets) == len(start_of_word)
```
Timoeller, line 319: here we either need tests or keep the assert?

farm/modeling/prediction_head.py
```diff
@@ -1503,7 +1508,6 @@ def merge_formatted_preds(preds_all):
 
         # This fn is used to align QA output of len=n_docs and Classification output of len=n_passages
         def chunk(iterable, lengths):
-            assert sum(lengths) == len(iterable)
```
Timoeller, line 1506: I do not know what this assert does... lets discuss in detail or throw a logger.error?

brandenchan:
See #468 for how these changes were actually implemented

PR #448, opened 2020-07-08
Upgrade transformers version

tholor:
Upgrading to latest transformers (3.0.2). This will allow us to move to the faster rust tokenizers and use Electra in Haystack.

### DIFF
--- a/farm/modeling/tokenization.py
+++ b/farm/modeling/tokenization.py
@@ -96,7 +96,7 @@ class Tokenizer:
         elif tokenizer_class == "XLMRobertaTokenizer":
             ret = XLMRobertaTokenizer.from_pretrained(pretrained_model_name_or_path, **kwargs)
         elif tokenizer_class == "RobertaTokenizer":
-            ret = RobertaTokenizer.from_pretrained(pretrained_model_name_or_path, add_prefix_space=True, **kwargs)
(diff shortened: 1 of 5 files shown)


tholor:
QA accuracy benchmark was running successfully

farm/modeling/tokenization.py
```diff
@@ -292,11 +292,10 @@ def _words_to_tokens(words, word_offsets, tokenizer):
         elif len(tokens) == 0:
             tokens_word = tokenizer.tokenize(w)
         else:
-            try:
+            if type(tokenizer) == RobertaTokenizer:
```
Timoeller, line 295: what about XLMRobertaTokenizer? What if we check if the \_\_name\_\_ string contains roberta to make it more general?

test/test_conversion.py
```diff
@@ -23,7 +23,7 @@ def test_conversion_adaptive_model(caplog):
 
 def test_conversion_inferencer(caplog):
     if caplog:
-        caplog.set_level(logging.CRITICAL)
+        caplog.set_level(logging.WARNING)
```
Timoeller, line 26: I would like to leave it at critical

tholor: Done

PR #449, opened 2020-07-09
Fix offset_end character for QA

brandenchan:
Fixes a big in QA regarding end character calculation. If the extracted span is followed by a whitespace, the end_ch used to be one greater than it should be

### DIFF
--- a/farm/modeling/predictions.py
+++ b/farm/modeling/predictions.py
@@ -181,7 +181,7 @@ class QACandidate:
         final_text = clear_text[start_ch: end_ch].strip()
         end_ch = start_ch + len(final_text)
 
-        return clear_text[start_ch: end_ch].strip(), start_ch, end_ch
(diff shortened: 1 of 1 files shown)


farm/modeling/predictions.py
```diff
@@ -177,6 +177,10 @@ def _span_to_string(self, token_offsets: List[int], clear_text: str):
             end_ch = len(clear_text)
         else:
             end_ch = token_offsets[end_t]
+
+        final_text = clear_text[start_ch: end_ch].strip()
+        end_ch = start_ch + len(final_text)
+
         return clear_text[start_ch: end_ch].strip(), start_ch, end_ch
```
tholor, line 184: Shall we change the return then to 
`return final_text, start_ch, end_ch` ?

brandenchan: Yes, definitely

Issue #457, opened 2020-07-13
Data Handler dataset creation fails if features are not computed for some baskets 


ftesser:
**Describe the bug**
method `farm.data_handler.processor.Processor._create_dataset` fails if features are not computed for some basket.
The specific case of the error message below is the training of SQuAD question answering, the context of the issue is the following:
 * the input SQuAD data contains some "errors" (some misalignment between answer text and selected text in the context);
 * this misalignment cause the exclusion of the features computation the `_featurize_samples` method (see the try-except):  https://github.com/deepset-ai/FARM/blob/b8b59c4c8fb98b92f24f0e92f9e4f44db5549323/farm/data_handler/processor.py#L295
 * finally in `_create_dataset` when the basket has no feature  `features_flat.extend`  https://github.com/deepset-ai/FARM/blob/b8b59c4c8fb98b92f24f0e92f9e4f44db5549323/farm/data_handler/processor.py#L308
gives the error message reported below.

**Error message**
```
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/home/fabio/src/git_repositories/FARM/farm/data_handler/data_silo.py", line 124, in _dataset_from_chunk
    dataset = processor.dataset_from_dicts(dicts=dicts, indices=indices)
  File "/home/fabio/src/git_repositories/FARM/farm/data_handler/processor.py", line 1144, in dataset_from_dicts
    dataset, tensor_names = self._create_dataset(keep_baskets=False)
  File "/home/fabio/src/git_repositories/FARM/farm/data_handler/processor.py", line 308, in _create_dataset
    features_flat.extend(sample.features)
TypeError: 'NoneType' object is not iterable
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/fabio/src/git_repositories/DocumentFeaturesIdentificator/bin/farm_qa", line 10, in <module>
    farm_qa.main()
  File "/home/fabio/src/git_repositories/DocumentFeaturesIdentificator/bin/../documentfeaturesidentificator/farm_utils/farm_qa.py", line 211, in main
    options.func(options)
  File "/home/fabio/src/git_repositories/DocumentFeaturesIdentificator/bin/../documentfeaturesidentificator/farm_utils/farm_qa.py", line 107, in train_general_purpose
[...]

PR #459, opened 2020-07-14
Remove baskets without features in  _featurize_samples.

ftesser:
This is proposals 1 for #457.
I have added some additional info on the id that contains errors (so if someone want to check the data can find them).
I am not aware if it is ok to remove baskets directly here with `self.baskets.remove(basket)`: any  contraindication for that?


ftesser:
This PR is not necessary anymore (it is substituted by https://github.com/deepset-ai/FARM/pull/471)

PR #460, opened 2020-07-14
Ensure QAInferencer always has task_type "question_answering"

brandenchan:
Though the parent Inferencer class can set task_type to one of many values, it should always be "question_answering" in the QAInferencer. This PR enforces this and gives a warning when a user tries to initialise the class with a different value

### DIFF
--- a/farm/infer.py
+++ b/farm/infer.py
@@ -630,7 +630,7 @@ class QAInferencer(Inferencer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.task_type != "question_answering":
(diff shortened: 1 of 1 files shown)


farm/infer.py
```diff
@@ -627,6 +627,12 @@ def extract_vectors(self, dicts, extraction_strategy="cls_token", extraction_lay
 
 
 class QAInferencer(Inferencer):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.task_type != "question_answering":
+            logger.warning("QAInferencer always has task_type='question_answer' even if another value is provided "
```
Timoeller, line 633: "question_answering" instead of 'question_answer'

PR #461, opened 2020-07-14
Remove baskets without features in _create_dataset.

ftesser:
This is proposals 2 for #457.
I have added some additional info on the id that contains errors (so if someone want to check the data can find them).

The removal is done on _create_dataset.

### DIFF
--- a/farm/data_handler/processor.py
+++ b/farm/data_handler/processor.py
@@ -299,13 +299,25 @@ class Processor(ABC):
                     sample.features = self._sample_to_features(sample=sample)
                 except Exception as e:
                     logger.error(f"Could not convert this sample to features: \n {sample}")
+                    logger.error(f"Basket id: id_internal: {basket.id_internal}, id_external: {basket.id_external}")
                     logger.error(f"Error message: {e}")
 
     def _create_dataset(self, keep_baskets=False):
         features_flat = []
+        basket_to_remove = []
         for basket in self.baskets:
             for sample in basket.samples:
-                features_flat.extend(sample.features)
+                if sample.features is None:
+                    basket_to_remove.append(basket)
+                else:
+                    features_flat.extend(sample.features)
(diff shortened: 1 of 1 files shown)


farm/data_handler/processor.py
```diff
@@ -299,13 +299,25 @@ def _featurize_samples(self):
...
+                if sample.features is None:
+                    basket_to_remove.append(basket)
+                else:
+                    features_flat.extend(sample.features)
+        if len(basket_to_remove) > 0:
+            # if basket_to_remove is not empty remove the related baskets
+            logger.warning(f"Removing the following baskets because of errors in computing features:")
+            for basket in basket_to_remove:
+                logger.warning(f"Basket id: id_internal: {basket.id_internal}, id_external: {basket.id_external}")
+                self.baskets.remove(basket)
```
brandenchan, line 319: An error could be thrown here if a basket contains multiple samples where sample.features is None

farm/data_handler/processor.py
```diff
@@ -299,13 +299,25 @@ def _featurize_samples(self):
...
     def _create_dataset(self, keep_baskets=False):
         features_flat = []
+        basket_to_remove = []
         for basket in self.baskets:
             for sample in basket.samples:
-                features_flat.extend(sample.features)
+                if sample.features is None:
+                    basket_to_remove.append(basket)
+                else:
+                    features_flat.extend(sample.features)
```
brandenchan, line 313: If one set of sample.features is None, all sample.features from the same basket need to be thrown out otherwise we may have features which belong to a basket which has been removed. So we should probably first iterate over all samples in the basket to check if any have features = None. If there are, we remove all features. If not, then we can add all to features_flat

ftesser: thanks @brandenchan for the suggestions. I am going to close this PR and create a new one.

PR #468, opened 2020-07-21
Remove assertions or replace with logging error

brandenchan:
This PR either removes asserts or replaces them with logging errors. The idea behind this is to improve the stability of FARM in production so that it doesn't crash halfway through a long inference or training job. Asserts that check initialization of components are maintained.


brandenchan:
See #444 for some earlier discussions around the topic

PR #474, opened 2020-07-27
Remove rest_api_schema argument in inference_from_dicts()

brandenchan:
Removes traces of an old argument that could be passed in to Inferencer.inference_from_dicts()

### DIFF
--- a/examples/conversion_huggingface_models.py
+++ b/examples/conversion_huggingface_models.py
@@ -22,7 +22,7 @@ def convert_from_transformers():
     # run predictions
     QA_input = [{"questions": ["Why is model conversion important?"],
                  "text": "The option to convert models between FARM and transformers gives freedom to the user and let people easily switch between frameworks."}]
-    result = nlp.inference_from_dicts(dicts=QA_input, rest_api_schema=True)
+    result = nlp.inference_from_dicts(dicts=QA_input)
     pprint.pprint(result)
     nlp.close_multiprocessing_pool()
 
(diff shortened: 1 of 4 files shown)


farm/inference_rest_api.py
```diff
@@ -84,7 +84,7 @@ def post(self, model_id):
         dicts = request.get_json().get("input", None)
         if not dicts:
             return {}
-        results = model.inference_from_dicts(dicts=dicts, rest_api_schema=True)
+        results = model.inference_from_dicts(dicts=dicts)
         return results[0]
```
Timoeller, line 88: results are not a list of QApreds, right_? Is this the right way to return those?

examples/conversion_huggingface_models.py
```diff
@@ -22,7 +22,7 @@ def convert_from_transformers():
     # run predictions
     QA_input = [{"questions": ["Why is model conversion important?"],
                  "text": "The option to convert models between FARM and transformers gives freedom to the user and let people easily switch between frameworks."}]
-    result = nlp.inference_from_dicts(dicts=QA_input, rest_api_schema=True)
+    result = nlp.inference_from_dicts(dicts=QA_input)
```
Timoeller, line 25: once we are at it, lets make nlp of type QAInferencer here

farm/inference_rest_api.py
brandenchan: rest_api_schema has been replaced by return_json. By default, return_json is True so we will get a dict output here instead of QAPreds

Timoeller: Ah ok thanks for clarification.

PR #500, opened 2020-08-25
Download models from (private) S3

tholor:
Simple utility function to download a model from an s3 bucket (e.g. private AWS or on-prem deployment).
We'll skip those files that have already been downloaded before.

Usage: 

```python
from farm.modeling.tokenization import Tokenizer
from farm.modeling.language_model import LanguageModel
from farm.file_utils import download_from_s3

# download model (if no custom cache_dir is supplied => we use FARM default cache dir, i.e. ~/.cache/torch/farm on Linux)
remote_model_path = "s3://your_bucket/bert-base-german-cased/"
local_model_path = download_from_s3(s3_url=remote_model_path, cache_dir=None)

# load components as usual
tokenizer = Tokenizer.load(local_model_path)
language_model = LanguageModel.load(local_model_path)
```

Potential improvements in the future:
- add a file with the hash to see if remote files have been updated and therefore need to be downloaded again
- progress bar for download

### DIFF
--- a/farm/file_utils.py
+++ b/farm/file_utils.py
@@ -105,11 +105,14 @@ def download_from_s3(s3_url: str, cache_dir: str = "~/.cache/torch/farm"):
     :return: local path of the folder
     """
 
+    logger.info(f"Downloading from {s3_url}")
     s3_resource = boto3.resource('s3')
     bucket_name, s3_path = split_s3_path(s3_url)
     bucket = s3_resource.Bucket(bucket_name)
     objects = bucket.objects.filter(Prefix=s3_path)
-    logger.info(f"Downloading from {s3_url}")
+    if not objects:
+        raise ValueError("Could not find s3_url: {s3_url}")
+
     for obj in objects:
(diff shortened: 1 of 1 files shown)


farm/file_utils.py
```diff
@@ -95,6 +96,36 @@ def filename_to_url(filename, cache_dir=None):
...
+    :param cache_dir: local directory where the files shall be stored
+    :return: local path of the folder
+    """
+
+    s3_resource = boto3.resource('s3')
+    bucket_name, s3_path = split_s3_path(s3_url)
+    bucket = s3_resource.Bucket(bucket_name)
+    objects = bucket.objects.filter(Prefix=s3_path)
+    logger.info(f"Downloading from {s3_url}")
+    for obj in objects:
```
tanaysoni, line 113: should we add a check here if objects exist?

tholor: Added an exception

tanaysoni (approved):

PR #503, opened 2020-08-26
Pass max_answers param to processor

brandenchan:
This propagates the max_answers param from the samples_to_features_qa() function up to the SquadProcessor and NaturalQuestionsProcessor so that they can be set in Haystack.


tholor (approved):
Looking good!

PR #504, opened 2020-08-27
Add warmup run to component benchmark

brandenchan:
We saw cases where the first run in the loop of question_answering_components.py took longer or showed different time profile characteristics. This PR adds a warm up run in order to combat this fluctuation, which we suspect comes from the GPU warming up.


tholor (approved):
Seems reasonable! Did you try running it and checked whether the timing is now more consistent?

tholor:
I did a test run and the timing seems now consistent :+1:

Issue #496, opened 2020-08-23
Store Test Eval. Result in Variable


PhilipMay:
FARM is doing test set evaluation when certain conditions are met. The results are only logged but it is impossible / hard to get them with downstream Python code. I think the test set evaluation results should be stored in a variable so they can be read. 

What do you think? As always I would crate a PR.

tholor:
Yes, I think it would be useful as long as it doesn't complicate the trainer too much or introduces breaking changes. 

Results could either be saved in a variable or potentially as an alternative also written to a file. The latter wouldn't introduce changes to the return signature of trainer.train(). Open do other design ideas, too.  

FYI There was a slightly related discussion in #353 about storing the **predictions** made during eval to a file (for external eval scripts). Possibly we could even tackle both use cases ...

PhilipMay:
I made a PR for this on #506 
It does not introduce any breaking change.

PR #506, opened 2020-08-30
Store test eval result in variable

PhilipMay:
- add self.test_result and init it with None
- set self.test_result with eval result
- add docstring
- write (extend) test

This does not introduce any breaking change.

Also see #496


tholor (approved):
Simple and effective. I like it! 
Thanks for adding this @PhilipMay !

PR #508, opened 2020-09-01
Add possibility to do nested cross validation

PhilipMay:
This adds the possibility to do nested cross validation.

The PR introduces a new class `DataSiloForNestedCrossVal` that inherits from `DataSiloForCrossVal`.
It does nested cross validation:

- other cross validation splits the data to a rest set and a test set
- inner cross validation splits the rest data to train and dev

This is related to #507

## Todo
- [x] write test
- [x] discuss with @Timoeller (and @johann-petrak) 
- [ ] ~~add quest. answering~~
- [x] add doc
- [x] add number of folds pram for outer and inner folds


Timoeller:
After some offline discussion with @PhilipMay we are going forward with the nested Crossvalidation. 

We need a way to handle the variance in performance that comes from different restarts (be it different seeds or slightly different training data). 

This PR adds a solution in a structured way, so glad to have it in FARM.

PhilipMay:
An other plus for this nested cross validation is that the train and dev datasets are also created in a stratified way.

PhilipMay:
Rebased on master.

test/data_handler/test_data_silo.py
```diff
@@ -0,0 +1,70 @@
+from pathlib import Path
```
Timoeller, line 1: lets leave this fct in the top level folder with all the other tests.
If we restructure the test folder we will do so in a separate PR and for all tests

test/data_handler/test_data_silo.py
```diff
@@ -0,0 +1,70 @@
...
+import numpy as np
+
+from farm.modeling.tokenization import Tokenizer
+from farm.data_handler.processor import TextClassificationProcessor
+from farm.data_handler.data_silo import DataSilo, DataSiloForNestedCrossVal
+
+
+def test_data_silo_for_nested_cross_val():
+    lang_model = "bert-base-german-cased"
+    n_outer_splits = 7
```
Timoeller, line 11: lets use less splits, otherwise processing takes longer than necessary. Or is there a special reason for these numbers?

farm/data_handler/data_silo.py
```diff
@@ -942,3 +942,104 @@ def _split_for_qa(documents, id_index, n_splits=5, shuffle=True, random_state=No
...
+        :param random_state: random state for shuffling
+        :type random_state: int
+        :param stratified: if class stratification should be done
+        :type stratified: bool
+        :param n_neg_answers_per_question: number of negative answers per question to include for training
+        :type n_neg_answers_per_question: int
+        """
+
+        if "question_answering" in datasilo.processor.tasks:
+            return cls._make_question_answering(datasilo, sets, n_splits, shuffle, random_state, n_neg_answers_per_question)
```
Timoeller, line 996: n_splits is not defined and will throw an error. We should nevertheless try to integrate the inner split functionality into DataSiloForCrossVal

farm/data_handler/data_silo.py
```diff
@@ -942,3 +942,104 @@ def _split_for_qa(documents, id_index, n_splits=5, shuffle=True, random_state=No
             current_train_set = np.hstack(np.delete(splits, idx, axis=0))
 
             yield current_train_set, current_test_set
+
+
+class DataSiloForNestedCrossVal(DataSiloForCrossVal):
```
Timoeller, line 947: Now that I see the code there is a lot of duplicate code in this class and DataSiloForCrossVal.
Cant we merge both classes and parameterize it?

farm/data_handler/data_silo.py
```diff
@@ -942,3 +942,104 @@ def _split_for_qa(documents, id_index, n_splits=5, shuffle=True, random_state=No
...
+    """
+    Prepare data silos to do a nested cross validation.
+
+    The outer cross validation splits all data to test and rest.
+    The inner cross validation splits the rest data to train and dev.
```
Timoeller, line 952: Lets motive the inner split in the docstring by talking about variance in multiple restarts

Timoeller (changes requested):
This is looking well but could be a bit better integrated into existing code.
I made some comments along the way - open to discussion

test/data_handler/test_data_silo.py
PhilipMay: done.

test/data_handler/test_data_silo.py
PhilipMay: yes - done

farm/data_handler/data_silo.py
PhilipMay: fixed

farm/data_handler/data_silo.py
PhilipMay: I have done that. @Timoeller what do you think about this solution?

farm/data_handler/data_silo.py
PhilipMay: I added extensive documentation. :-)

PhilipMay:
@Timoeller I made your suggested changes.
I hope the integration into the existing class is fine with you.

I had a look on nested cross validation for question answering.
But to be honest the whole thing is a bit too fiddly for me. 
I'm not deep enough in the question answering topic and would build many bugs I guess.

Maybe you could just add this later when you / someone else wants / needs it?

IMO this is good to go to be merged.

Timoeller (approved):
Looking good, thanks for integrating the proposed changes.

PR #515, opened 2020-09-04
Bump transformers version to 3.1.0

tholor:


tholor:
No, I thought we only do that before any release

Timoeller (approved):
True, I still did the accuracy check and it passed all test.

Issue #516, opened 2020-09-04
FastTokenizer gives different output for sequence starting with a subword token


tholor:
**Describe the bug**
In QA we split long texts into multiple passages with a sliding window after tokenization. 
This means a long word with multiple tokens might be splitted across two passages (start of word in passage 1, rest in passage 2). For these cases, the current implementation of fast tokenizer will produce different results than the slow tokenizer. 

```
[...]

PR #518, opened 2020-09-04
WIP fix passage split in the middle of a word

tholor:
Addressing #516


tholor:
Will be fixed in #624

PR #521, opened 2020-09-05
fix Natural Question inference processing

Timoeller:
When extracting IDs from the input we run into errors.

This is a quick fix catching the exception, since we do not need these IDs. 
We should investigate for a better solution and create NQ test cases for training and inferencing @brandenchan

### DIFF
--- a/examples/natural_questions.py
+++ b/examples/natural_questions.py
@@ -35,7 +35,7 @@ def question_answering():
     evaluate_every = 500
     lang_model = "deepset/roberta-base-squad2" # start with a model that can already extract answers
     do_lower_case = False # roberta is a cased model
-    train_filename = "train_medium.jsonl"
+    train_filename = "dev_medium.jsonl"
     dev_filename = "dev_medium.jsonl"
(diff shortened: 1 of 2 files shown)


brandenchan:
Yup this quick fix is sufficient for now. There will be a bigger refactor coming where we merge the SQuAD and NQ processors. There we can more accurately define the supported inference input. We should also write tests to ensure that each inference input format works.

examples/natural_questions.py
```diff
@@ -35,7 +35,7 @@ def question_answering():
     evaluate_every = 500
     lang_model = "deepset/roberta-base-squad2" # start with a model that can already extract answers
     do_lower_case = False # roberta is a cased model
-    train_filename = "train_medium.jsonl"
+    train_filename = "dev_medium.jsonl"
```
brandenchan, line 38: Let's revert this back to train_medium.jsonl

Timoeller:
k, thanks for the input. Merging this now as fix.

brandenchan (changes requested):
Apart from changing the input file, this all looks good to me!

Issue #522, opened 2020-09-06
Trainer does not use loaded model for test set evaluation


PhilipMay:
When using early stopping the best models are saved. Later they are loaded here:
https://github.com/deepset-ai/FARM/blob/c6871b67401eef36f9d4310c966063eda10b0dfe/farm/train.py#L355-L356

But here the evaluator does not get the loaded model `model` but the old model `self.model`:
https://github.com/deepset-ai/FARM/blob/c6871b67401eef36f9d4310c966063eda10b0dfe/farm/train.py#L365

This is a bug...

PhilipMay:
Also the question is if the trainer should return the last model or the best model? See here:

https://github.com/deepset-ai/FARM/blob/c6871b67401eef36f9d4310c966063eda10b0dfe/farm/train.py#L367

PhilipMay:
Fix in #524

PhilipMay:
PR was merged. Closing this.

Timoeller:
Thanks for spotting and also fixing!

PhilipMay:
Was a nasty little bug - right? :-)

PR #523, opened 2020-09-06
Evaluate model on best model when doing early stopping

PhilipMay:
Fix for #522


PhilipMay:
This is based on wrong branch. Closing again.

PR #527, opened 2020-09-07
auto lowercase: set lower case based on model name

scheiblr:
added the possibility to let the framework estimate lowercase modules from the module name. I observed the unwritten standard, that lowercase modules contain `uncased` in the name. In order to mix uncased and cased in one experiment file, this feature was added. Simply set `lower_case` to `"auto"` in the experiment json file.

### DIFF
--- a/farm/data_handler/processor.py
+++ b/farm/data_handler/processor.py
@@ -197,7 +197,17 @@ class Processor(ABC):
         if "lower_case" in config.keys():
             logger.warning("Loading tokenizer from deprecated FARM config. "
                            "If you used `custom_vocab` or `never_split_chars`, this won't work anymore.")
-            tokenizer = Tokenizer.load(load_dir, tokenizer_class=config["tokenizer"], do_lower_case=config["lower_case"])
+                           
+            # automatically estimate lowercase model based on the model name
+            if config["lower_case"] == 'auto':
+                if 'uncased' in config["model"]:
+                    lower_case=True
+                else:
+                    lower_case=False
+            else:
+                lower_case = config["lower_case"]
+
+            tokenizer = Tokenizer.load(load_dir, tokenizer_class=config["tokenizer"], do_lower_case=lower_case)
         else:
             tokenizer = Tokenizer.load(load_dir, tokenizer_class=config["tokenizer"])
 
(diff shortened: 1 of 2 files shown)


Timoeller:
Thanks for the PR. I like the idea of automatically detecting the lowercasing info.
Wrong lower casing results in unexpected behavior. Especially models with upper case info can deal with uncased input as well, often resulting in just minor performance differences that are hard to spot.

How about we implement a more general approach, since this PR only adds changes made to the experiments section. I would prefer to completely get rid of the lowercasing variable in all cases of FARM.
Would you like to work on this @scheiblr? We can support you if you do.

scheiblr:
Would you have an idea of how this could be approached with a generic solution?

Timoeller:
Might be difficult to cover all cases. What about looking at the config files that are loaded with the models? Modern models from HF modelhub and FARM models should have the necessary information in there...

scheiblr:
Maybe, a mixed approach could be implemented: checking config and if information is not available, check model name (as I did), take some default else. What do you think?

Timoeller:
Nice, that would be a good way forward.

I would still like to have a lowercasing parameter present, so people can decide. The default val should then be "auto" as you suggested.

scheiblr:
Yes, I think it's important to have the possibility of full control.
You want it be everywhere, not only for the experiments, right? Cause I dunno which parts of the library would be affected, too.

Btw.: I just looked at my config files, and there was no information about the casing.

Timoeller:
What do you mean? When a model is saved in FARM in the tokenizer config there is do_lower_case information.

For transformers it is different, old models do not have casing information, newer models should have. Might be worth looking into the autotokenizers from transformers and how they do it.

Timoeller:
Hey any updates on this @scheiblr ?

Just to confirm terminology: there are some config files in our experiments folder, these are very different configs to the ones attached to a model. Models often have a tokenizer config, language model config and prediction head specific configs. The lowercasing information is within the tokenizer config of the languagemodel itself.

Timoeller:
Closing this PR for now since it seems stale. If you make any progress here please create a new one @scheiblr 

Thanks!

Issue #525, opened 2020-09-07
Warning message spam "please use `truncation=True` to explicitely truncate examples to max length"


PhilipMay:
Hi,

since I use the code from master branch I am getting spammed with these messages:

```
Truncation was not explicitely activated but `max_length` is provided a specific value, please use `truncation=True` to explicitely truncate examples to max length. Defaulting to 'longest_first' truncation strategy. If you encode pairs of sequences (GLUE-style) with the tokenizer you can select this strategy more precisely by providing a specific strategy to `truncation`.
```

I think this is related to #482 .

tholor:
Hey, 
For which task do you get this? Are you using fast or slow tokenizers?

PhilipMay:
Text classification. Since I did not specify to use fast I think I use slow.

tholor:
Fixed in #528

PR #528, opened 2020-09-07
Fix truncation warnings in tokenizer

tholor:
Solving #525

### DIFF
--- a/farm/data_handler/input_features.py
+++ b/farm/data_handler/input_features.py
@@ -43,6 +43,8 @@ def sample_to_features_text(
         # This should change once we git rid of FARM's tokenize_with_metadata()
         inputs = tokenizer(text,
                            return_token_type_ids=True,
+                           truncation=True,
+                           truncation_strategy="longest_first",
                            max_length=max_seq_len,
                            return_special_tokens_mask=True)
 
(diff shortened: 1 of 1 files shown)


farm/data_handler/input_features.py
```diff
@@ -43,6 +43,8 @@ def sample_to_features_text(
         # This should change once we git rid of FARM's tokenize_with_metadata()
         inputs = tokenizer(text,
                            return_token_type_ids=True,
+                           truncation=True,
+                           truncation_strategy="longest_first",
```
Timoeller, line 47: Why do we want to have truncation=True when we set it to false in line 65?
Also truncation_strategy is deprecated and has been removed in earlier commits. Is it necessary now?

Are we not doing truncation outside of encoding?

Timoeller (commented):
Posed some questions as comment

farm/data_handler/input_features.py
tholor: For the fast tokenizer we feed here the whole untruncated text again (therefore we need truncation). In line 65 we use the slow tokenizer (which just gets the truncated tokens)

Timoeller (approved):
looking good

PhilipMay:
Thanks. :-)

PR #532, opened 2020-09-08
Update torch requirement to make it compatible with Windows OS

tanaysoni:
Related issue: https://github.com/deepset-ai/haystack/issues/360

### DIFF
--- a/readme.rst
+++ b/readme.rst
@@ -137,6 +137,12 @@ From PyPi::
 
     pip install farm
 
+
+Windows OS:
+
+For installation on Windows, add :code:`-f https://download.pytorch.org/whl/torch_stable.html` to the pip install.
+
+
 Basic Usage
 ############
 1. Train a downstream model
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ setuptools
 wheel
 # PyTorch
 --find-links=https://download.pytorch.org/whl/torch_stable.html
-torch==1.6.0
+torch>=1.5.1,<=1.6.0
 # progress bars in model download and training scripts
 tqdm
 # Accessing files from S3 directly.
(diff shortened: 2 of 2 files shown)


tholor:
Is windows not supporting ranges for pip?
It's not ideal to always pin a single torch version here as many users are still using previous versions (which are totally compatible with haystack).

Timoeller:
The special >=xxx,<=xxx syntax as well as the * operator does not work for pip under windows.

How about we just have `torch>=1.5.1`?

tholor:
I am not sure if that's really the case. 
I would really try to avoid fixing a version here. 
Other libs like spacy use ranges, too (https://github.com/explosion/spaCy/blob/master/requirements.txt)
Maybe only a pytorch specific thing? 
Let's test this on a windows machine ...

PiffPaffM:
The installation process seems to work if we install the requirements via the requirements.txt file (`pip install -r requirements.txt`) If we install it via `pip install --editable .`, the error occurs. 
In the requirements.txt file is the following command used which might be the reason for this behaviour: 
`--find-links=https://download.pytorch.org/whl/torch_stable.html`

tholor:
After testing with @PiffPaffM, we found: 
- It's not an issue with the version range
- It's due to pytorch not having windows versions on pypi
- Therefore, pytorch installation on windows relies on `--find-links=https://download.pytorch.org/whl/torch_stable.html` and is also recommended on their website

Unfortunately, installation via pip doesn't allow an equivalend"find-links
Recommended solution: 
- let's keep the version range 
- add option for win install to docs: `pip install farm-haystack -f https://download.pytorch.org/whl/torch_stable.html`

tholor:
I already updated the readme. I think we can close this PR

Zenahr:
@tholor to make it windows compatible we would have to remove `uvloop` from the requirements as well.

stack trace:
```
raise RuntimeError('uvloop does not support Windows at the moment')
    RuntimeError: uvloop does not support Windows at the moment
```

Issue #536, opened 2020-09-13
The `doc_stride` Parameter in `chunk_into_passages` Can Cause Errors or Unexpected Behaviour


EhsanM4t1qbit:
**Describe the bug**
I suspect that there is a bug in the function `chunk_into_passages` in `samples.py`, used for breaking down a long paragraph into multiple passages for QA tasks.
There is a moving window for selecting a chunk of the paragraph. The window starting point is `passage_start_t` which moves by `doc_stride` tokens, while the window end token, `passage_end_t`, moves by `passage_len_t` tokens. I see a few problematic possible scenarios here. 

```
passage_id = 0
doc_len_t = len(doc_offsets)    
while True:
    passage_start_t = passage_id * doc_stride
    passage_end_t = passage_start_t + passage_len_t
    passage_start_c = doc_offsets[passage_start_t].   #  line 228
.
.
.
    if passage_end_t >= doc_len_t:
        break   
```

- `doc_stride` > `doc_len_t` > `passage_len_t`: This will cause an error on line 228.
```
E.g. doc_stride = 200, doc_len_t = 150, passage_len_t = 100
First passage: tokens 0-100
Second passage: tokens 200-300


09/11/2020 19:23:15 - ERROR - farm.data_handler.processor -   Error message: list index out of range
```

- `doc_len_t` > `doc_stride` >  `passage_len_t`: This will silently skip a number of tokens. 
```
E.g. doc_len_t = 200, doc_stride = 150, passage_len_t = 100
First passage: tokens 0-100
Second passage: tokens 150-250
```

- `doc_stride` < `passage_len_t`: There will be an overlap between the two chunks.
```
E.g. doc_len_t = 200, doc_stride = 100, passage_len_t = 150
First passage: tokens 0-150
Second passage: tokens 100-250
```

Note that it's not straightforward to set `passage_len_t` since it is dependent on a number of other parameters.
```
passage_len_t = max_seq_len - question_len_t - n_special_tokens
```
The simple solution is to get rid of `doc_stride` and set `passage_start_t` to `passage_end_t+1` at the end of the while loop.

brandenchan:
Hey @EhsanM4t1qbit, thanks for the issue. This seems very related to #510 and I totally agree we need to have some kind of message to avoid situations where doc_stride > passage_len_t. This has been introduced with #538

[...]

Issue #510, opened 2020-09-01
Add assert for QAInferencer's max_seq_len & doc_stride


tanaysoni:
The `QAInferencer` throws an error when the `max_seq_len` is less than the `doc_stride`. 

Here's the stack trace:
```python
  File "../haystack/haystack/finder.py", line 61, in get_answers
    top_k=top_k_reader)  # type: Dict[str, Any]
  File "../haystack/haystack/reader/farm.py", line 324, in predict
    objects=inputs, return_json=False, multiprocessing_chunksize=1
  File "../.pyenv/versions/3.7.6/envs/haystack/lib/python3.7/site-packages/farm/infer.py", line 699, in inference_from_objects
    multiprocessing_chunksize=multiprocessing_chunksize, streaming=streaming)
  File "../.pyenv/versions/3.7.6/envs/haystack/lib/python3.7/site-packages/farm/infer.py", line 682, in inference_from_dicts
    multiprocessing_chunksize=multiprocessing_chunksize, streaming=streaming)
  File "../.pyenv/versions/3.7.6/envs/haystack/lib/python3.7/site-packages/farm/infer.py", line 433, in inference_from_dicts
    predictions = self._inference_without_multiprocessing(dicts, return_json, aggregate_preds)
  File "../.pyenv/versions/3.7.6/envs/haystack/lib/python3.7/site-packages/farm/infer.py", line 477, in _inference_without_multiprocessing
    dicts, indices=[i for i in range(len(dicts))], return_baskets=True
  File "../.pyenv/versions/3.7.6/envs/haystack/lib/python3.7/site-packages/farm/data_handler/processor.py", line 1166, in dataset_from_dicts
    self._featurize_samples()
  File "../.pyenv/versions/3.7.6/envs/haystack/lib/python3.7/site-packages/farm/data_handler/processor.py", line 297, in _featurize_samples
    for sample in basket.samples:
TypeError: 'NoneType' object is not iterable
```

<br>
It might also be nice to have assertions on other parameters that they have values within a reasonable range.

tholor:
Fixed with #538

PR #538, opened 2020-09-14
Add asserts on doc_stride and max_seq_len to prevent issues with sliding window

brandenchan:
In cases where the doc_stride is greater than the number of passage tokens in the sliding window, an index error may be thrown. See #536 and #510. This PR adds asserts to avoid these situations.  

Note that the number of passage tokens is computed as follows:
max_seq_len - question_tokens - special tokens

### DIFF
--- a/farm/data_handler/samples.py
+++ b/farm/data_handler/samples.py
@@ -219,6 +219,11 @@ def chunk_into_passages(doc_offsets,
                         doc_text):
     """ Returns a list of dictionaries which each describe the start, end and id of a passage
     that is formed when chunking a document using a sliding window approach. """
+
+    assert doc_stride < passage_len_t, "doc_stride is longer than passage_len_t. This means that there will be gaps " \
+                                       "as the passage windows slide, causing the model to skip over parts of the document. "\
+                                       "Please set a lower value for doc_stride (Suggestions: doc_stride=128, max_seq_len=384) "
+
     passage_spans = []
     passage_id = 0
     doc_len_t = len(doc_offsets)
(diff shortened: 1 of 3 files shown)


farm/data_handler/samples.py
```diff
@@ -219,6 +219,11 @@ def chunk_into_passages(doc_offsets,
                         doc_text):
     """ Returns a list of dictionaries which each describe the start, end and id of a passage
     that is formed when chunking a document using a sliding window approach. """
+
+    assert doc_stride < passage_len_t, "doc_stride is longer than passage_len_t. This means that there will be gaps " \
```
tholor, line 223: Let's switch this one to a logger.error() as it can be thrown at runtime rather than at init (like the other two)

brandenchan: All errors thrown during the creation of samples and features are caught by a higher level try-catch (see processor._init_samples_in_baskets) which is why I thought I'd go for an assert statement here. I think it will be clearer to the user if its thrown as an Error here, then caught and printed at the higher level. Would you agree?

tholor: Ok makes sense

tholor (approved):

farm/data_handler/samples.py
Timoeller: This more localized but caught assert I would prefer.

What about a more user friendly error msg since the user doesnt know about passage_len_t. 
What about describing passage len t: "Doc stride is longer than max_seq_len minus the tokens used for the question"

Timoeller (approved):
I think it is good for merging, just made a comment to the caught exception.

Issue #529, opened 2020-09-07
`examples/natural_questions.py` evaluation not computed for yes/no classes 


ftesser:
**Describe the bug**
Running the `examples/natural_questions.py` script I obtain the following evalution:

```
\\|//       \\|//      \\|//       \\|//     \\|//
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
***************************************************
***** EVALUATION | DEV SET | AFTER 2000 BATCHES *****
***************************************************
\\|//       \\|//      \\|//       \\|//     \\|//
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

09/07/2020 11:56:01 - INFO - farm.eval -   
 _________ question_answering _________
09/07/2020 11:56:01 - INFO - farm.eval -   loss: 1.1128250570618547
09/07/2020 11:56:01 - INFO - farm.eval -   task_name: question_answering
09/07/2020 11:56:01 - INFO - farm.eval -   EM: 0.632
09/07/2020 11:56:01 - INFO - farm.eval -   f1: 0.6685447088801921
09/07/2020 11:56:01 - INFO - farm.eval -   top_n_accuracy: 0.918
09/07/2020 11:56:01 - INFO - farm.eval -   report: 
 Not Implemented
09/07/2020 11:56:01 - INFO - farm.eval -   
 _________ text_classification _________
09/07/2020 11:56:01 - INFO - farm.eval -   loss: 0.3979932257842103
09/07/2020 11:56:01 - INFO - farm.eval -   task_name: text_classification
09/07/2020 11:56:01 - INFO - farm.eval -   f1_macro: 0.8045831308172903
09/07/2020 11:56:01 - INFO - farm.eval -   report: 
               precision    recall  f1-score   support

   no_answer     0.8355    0.8806    0.8575       871
        span     0.7878    0.7188    0.7517       537
         yes     0.0000    0.0000    0.0000         0
          no     0.0000    0.0000    0.0000         0

   micro avg     0.8189    0.8189    0.8189      1408
   macro avg     0.4058    0.3999    0.4023      1408
weighted avg     0.8173    0.8189    0.8171      1408
```
In the text_classification part `yes` and `no` classes are reported as 0 for all the scores also support, but I checked that in the dev-set some `yes`/`no` classes are present.

**Expected behavior**
[...]

PR #540, opened 2020-09-14
Fix removal of yes no answers

brandenchan:
Previously, yes no answers in the Natural Questions dataset were being converted into span answers (#529). This PR fixes this problem.


farm/data_handler/processor.py
```diff
@@ -1744,10 +1745,13 @@ def _apply_tokenization(dictionary, tokenizer):
             external_id = question["id"]
             question_text = question["question"]
             for answer in question["answers"]:
-                if answer["text"] == "":
-                    answer_type = "no_answer"
+                if 'answer_type' in answer.keys() and answer['answer_type'] in ['yes', 'no', 'span', 'no_answer']:
```
Timoeller, line 1748: I am unhappy about having the answer types hard coded here as a list. We should somehow link the already defined answer_type_list attribute of NaturalQuestionsProcessor.

Timoeller (changes requested):
Lets remove hard coded answer types as suggested in comment.

farm/data_handler/processor.py
tholor: Can we finish this before the next release?

Timoeller:
Hey sorry @brandenchan but this is not the best solution. We have an empty list as default arg which is an anti pattern, see https://docs.python-guide.org/writing/gotchas/

Otherwise the solution is ok with me.

PR #541, opened 2020-09-14
Fix logging of error msg for FastTokenizer + QA

tholor:
We were accessing a wrong key here resulting in an error within the error logging ;)


Timoeller (approved):
Ok

PR #542, opened 2020-09-14
Make use of the answer_type linked to each answers.

ftesser:
This PR is linked to issue #529 
This PR take care of the generic cases where there are more than one answers for each question.

### DIFF
--- a/examples/natural_questions.py
+++ b/examples/natural_questions.py
@@ -126,14 +126,16 @@ def question_answering():
     fetch_archive_from_http("https://s3.eu-central-1.amazonaws.com/deepset.ai-farm-qa/models/roberta-base-squad2-nq.zip", output_dir="../saved_models/farm")
     QA_input = [
         {
-            "qas": ["Did GameTrailers rated Twilight Princess as one of the best games ever created?"],
+            "qas": ["Did GameTrailers rated Twilight Princess as one of the best games ever created?", "Did GameTrailers rated Twilight Princess as one of the worst games ever created?"],
             "context":  "Twilight Princess was released to universal critical acclaim and commercial success. It received perfect scores from major publications such as 1UP.com, Computer and Video Games, Electronic Gaming Monthly, Game Informer, GamesRadar, and GameSpy. On the review aggregators GameRankings and Metacritic, Twilight Princess has average scores of 95% and 95 for the Wii version and scores of 95% and 96 for the GameCube version. GameTrailers in their review called it one of the greatest games ever created."
         }
     ]
 
     model = QAInferencer.load(model_name_or_path="../saved_models/farm/roberta-base-squad2-nq", batch_size=batch_size, gpu=True)
-    result = model.inference_from_dicts(dicts=QA_input, return_json=False) # result is a list of QAPred objects
 
+    result = model.inference_from_dicts(dicts=QA_input, return_json=False) # result is a list of QAPred objects
+    print(result)
+    print(len(result))
     print(f"\nQuestion: Did GameTrailers rated Twilight Princess as one of the best games ever created?"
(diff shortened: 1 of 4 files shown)


brandenchan:
So it's really cool that you’re getting so deep into the code now! We really appreciate that you're suggesting solutions and creating PRs. 

So just setting the scene, each dev question in NQ has 5 annotations. Each on of these can be of answer_type [yes, no, long_answer, short_answer, no_answer]. 

I totally agree that each annotation should be paired with an answer_type. But this is not how it is implemented right now since this part of the code is also shared with our SQuAD system. At the moment, all annotations of the same question have the same answer type, more specifically the answer_type of the first annotation, (see [this](https://github.com/deepset-ai/FARM/blob/36850354b1f5fe88f2b33a0c2d954e4f20399953/farm/data_handler/processor.py#L1435)). Because of this, I don’t believe that your PR will actually do what you’re intending sadly. But we agree that this needs to be fixed.

As it stands, the evaluation numbers that you see during training will be worse than they should be. (As an aside, you should also know that these evaluation numbers are calculated on a token level instead of a word level). But training and inference should be unaffected by this, meaning that you could still train, perform inference and evaluate using the official nq_eval.py script.

This part of the code will be significantly refactored in the coming weeks. It’s definitely on our roadmap. If time is pressing for you, and you would like to have accurate eval numbers during training, we can definitely give you pointers to a hacky solution. Just open up a new issue! But for now we’ll be closing this PR.

ftesser:
Thanks for the explanation, I agree that this change is not coherent with the actual FARM NQ process.

I am interesting in having multiple answers from the same question. 
The problem  that i would like to solve is not to work with the NQ dataset, but just allow a "normal" squad QA to answer also yes/no. So I was thinking that having your dictionary structure that is essentiualy in squad format + yes/no was a very good starting point.

To do this I would like to:

1) Allow to load squad json file where the answer can have the usual  "answer_start" and "text" fields with the addition of answer_type. 

2) Like in the squad dateset I would like to have the possibility of having  more than one answer for the same questions.
  
Point 2) was the real reason of my PR, but I see that is not actually ok with the other preprocessing of NQ dataset.


Could be FARM be interested in supporting something like this (json squad style + answer_type)? In that case I can try to write some code and make a PR.

Timoeller:
Thanks also from my side for digging deep into FARM and QA handling.

We will leave this PR open as a gentle reminder to ourselves to work on the NQ dataprocessing in on of our next sprints. Is that OK with you @ftesser or do you need progress on this task more immediately?

ftesser:
Thanks @Timoeller and all the FARM people for the FARM project.

Referring these last changes above (8 commits): they are not totally related to issue #529: probably it is better that I will make some single PR for each specif commit. What do you think?

Referring to work on the NQ data processing I have not time-constrains, just for curiosity: when do you think to place this sprint?

Timoeller:
We did include yes no answer type in #540 
Can we close this PR then @ftesser ?

ftesser:
In this PR, there is something that is not in the scope of issue #529.

In particular:

* Adding the possibility of referring to different heads of metrics for (this is useful for NaturalQuestions for tuning the earlystopping into class metrics instead of squad metrics)
* Return eval_value also when min_evals is not reached. is a bug I discovered into earlystopping

If you like I can open separate issue and PR for these, in this case yes you can close this PR.

Timoeller:
Hey thanks @ftesser for pointing this out. I can see that the changes in class EarlyStopping might be useful, but there is much other unmergable code in this PR as well.

Please create a separate PR with the EarlyStopping changes. Those changes would need to be thoroughly handled and tested though - currently I do not see any changes where instances of EarlyStopping are actually used (there is a different return val that needs be handled).

Closing this PR in favor of a more to Earlystopping tailored one.

Issue #533, opened 2020-09-10
Error in conversion of BertForMaskedLM with BertLMHead to HuggingFace transformers


himanshurawlani:
**Describe the bug**
I started with the [lm_finetuning.py](https://github.com/deepset-ai/FARM/blob/master/examples/lm_finetuning.py) script to pre-train BERT on a domain-specific text corpus. After completing training, the saved artifacts included the following:
* language_model.bin
* language_model_config.json (BertForMaskedLM)
* prediction_head_0.bin
* prediction_head_0_config.json (BertLMHead)
* prediction_head_1.bin
* prediction_head_1_config.json (NextSentenceHead)
* processor_config.json
* special_tokens_map.json
* tokenizer_config.json
* vocab.txt

I know FARM currently doesn't support converting an AdaptiveModel with two prediction heads to a HuggingFace transformers model. However, after going through the [AdaptiveModel.load()](https://github.com/deepset-ai/FARM/blob/d7b5cb61f062436789aa9081a85ade4f56e2db6f/farm/modeling/adaptive_model.py#L305) classmethod used in the model conversion [script](https://github.com/deepset-ai/FARM/blob/master/examples/conversion_huggingface_models.py), I figured out that removing the `prediction_head_1_config.json` (NextSentenceHead) and `prediction_head_1.bin` (PyTorch model) files should result in successful conversion to HuggingFace model as the conversion script would consider it as an AdaptiveModel with single PredictionHead.

**Error message**
After removing the files and running the conversion script, I get an error saying:
```
Traceback (most recent call last):
  File "conversion_huggingface_models.py", line 88, in <module>
    convert_to_transformers("./farm_saved_models/bert-english-lm", 
  File "conversion_huggingface_models.py", line 46, in convert_to_transformers
    transformer_model = model.convert_to_transformers()
  File "/home/himanshu/.conda/envs/tf2/lib/python3.8/site-packages/farm/modeling/adaptive_model.py", line 509, in convert_to_transformers
    elif len(self.prediction_heads[0].layer_dims) != 2:
  File "/home/himanshu/.conda/envs/tf2/lib/python3.8/site-packages/torch/nn/modules/module.py", line 771, in __getattr__
    raise ModuleAttributeError("'{}' object has no attribute '{}'".format(
torch.nn.modules.module.ModuleAttributeError: 'BertLMHead' object has no attribute 'layer_dims'
```
[...]

PR #555, opened 2020-09-23
Fixed conversion of BertForMaskedLM to transformers

himanshurawlani:
Fix for issue #533 

Made the following changes:

1. Added `layer_dims` attribute in FARM's BertLMHead class as described in the issue.
2. Added check for the conversion of AdaptiveModel with 2 prediction heads (where the first prediction head is "language_modelling") in `AdaptiveModel.convert_to_transformers()`


Timoeller (approved):
Looking good, thanks for this simple yet powerful addition.

Issue #539, opened 2020-09-14
Move ONNX conversion to transformers


Timoeller:
### Currently FARM onnx conversion does not support all model types. 
More specifically FARM does not adjust the naming and amount of input tensors to a model when converting to onnx. Implementing this for each model type can be cumbersome.

### Transformers ONNX conversion
Transformers now has a onnx conversion functionality for their pipeline models (example notebook [here](https://github.com/huggingface/transformers/blob/master/notebooks/04-onnx-export.ipynb)). I tested the conversion for bert, roberta and xlm-roberta with success.
So lets focus more on FARM <> transformers interoperability like in [this FARM example](https://github.com/deepset-ai/FARM/blob/master/examples/conversion_huggingface_models.py) and refer people to their conversion script(s).

We will therefore soon deprecate FARMs ONNX conversion functionality

PR #557, opened 2020-09-24
Add ONNX conversion & Inference

tanaysoni:
This PR extends supports for ONNX model conversion & inference using the transformers package. (Issue: #539)

Currently, `BERT`, `RoBERTa`, and `XLM-RoBERTa` models are supported.


farm/modeling/adaptive_model.py
```diff
@@ -631,139 +628,47 @@ def convert_from_transformers(cls, model_name_or_path, device, task_type, proces
...
+
+        task_type_to_pipeline_map = {
+            "question_answering": "question-answering",
+            "embeddings": "feature-extraction",
+            "ner": "ner"
+        }
 
-        tokenizer = Tokenizer.load(
-            pretrained_model_name_or_path="deepset/bert-base-cased-squad2"
+        convert(
```
tholor, line 643: Does this convert function include the same onnx optimizations as we had before? We saw quite big improvements due to them in our benchmarks ..

tholor (approved):
Looking good. Only minor comments.

farm/modeling/adaptive_model.py
```diff
@@ -818,13 +725,21 @@ def forward(self, **kwargs):
...
+            if self.language_model_class == "Bert":
+                input_to_onnx = {
+                    'input_ids': numpy.ascontiguousarray(kwargs['input_ids'].cpu().numpy()),
+                    'attention_mask': numpy.ascontiguousarray(kwargs['padding_mask'].cpu().numpy()),
+                    'token_type_ids': numpy.ascontiguousarray(kwargs['segment_ids'].cpu().numpy()),
+                }
+            elif self.language_model_class == "Roberta":
+                input_to_onnx = {
+                    'input_ids': numpy.ascontiguousarray(kwargs['input_ids'].cpu().numpy()),
+                    'attention_mask': numpy.ascontiguousarray(kwargs['padding_mask'].cpu().numpy())
```
tholor, line 737: What is needed if we add a new model here? Just adding the mapping for input names?
If yes, shall we also add XLM-R as it was requested by some users in the community?

farm/modeling/adaptive_model.py
tanaysoni: Many optimizations are now part of ONNXRuntime and are applied automatically when an inference session is created. 

I added two additional optimizations:
1. Conversion to float16: Earlier we were using optimization scripts from `farm.conversion` but now they're available in the `onnxruntime_tools` Python package. The conversion now uses this package and a new parameter `convert_to_float16` is added in `convert_to_onnx()` to apply this optimization.
2. Quantization: An option is added for the quantization of floating-point numbers to integers provided by the transformers that can be used by setting the `quantize` parameter in `convert_to_onnx()`.

farm/modeling/adaptive_model.py
tanaysoni: Yes, just the mapping is needed. However, since XLMRoberta exceeds the default limit of export size in PyTorch, so the parameter `use_external_format` for `convert()` has to be set.

PR #560, opened 2020-09-28
Add passages per second calculation to benchmarks

brandenchan:
Running the per component benchmark now returns a measurement of passages processed per second

### DIFF
--- a/test/benchmarks/question_answering_components.py
+++ b/test/benchmarks/question_answering_components.py
@@ -25,8 +25,8 @@ params = {
     "modelname": ["deepset/bert-base-cased-squad2", "deepset/minilm-uncased-squad2", "deepset/roberta-base-squad2", "deepset/bert-large-uncased-whole-word-masking-squad2", "deepset/xlm-roberta-large-squad2"],
     "batch_size": [50],
     "document_size": [1000_000],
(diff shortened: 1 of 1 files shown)


test/benchmarks/question_answering_components.py
```diff
@@ -17,28 +18,30 @@
 sample_file = "samples/question_answering_sample.txt"
 questions_file = "samples/question_answering_questions.txt"
 num_processes = 1
+passages_per_char = 2400 / 1000000      # numerator is number of passages when 1mill chars paired with one of the questions, msl 384, doc stride 128
```
tholor, line 21: If I get this right, it depends on the `document_size` of 1_000_000 that we define below in `params`. 
If we change `document_size` below this calculation will be wrong. 

How about we add a constant here instead `avg_chars_per_token`? It would then depend mainly on text + a bit on tokenizer. We could calculate it from your numbers as 1_000_000 / (384*2400) = 1.08 
and in params we add `num_passages = document_size / (max_seq_len * avg_chars_per_token)`

What do you think?

tholor: Ok nevermind, this would not factor in the doc_stride. Let's stay with the simple solution for now

Issue #558, opened 2020-09-24
QA Answers at the Beginning of the Document are Labeled as  (0, 0)


EhsanM4t1qbit:
I suspect that there is a bug in the function `generate_labels` https://github.com/deepset-ai/FARM/blob/master/farm/data_handler/input_features.py#L577.  The conditional statement should be changed to `passage_len > start_idx >= 0`. In its current form, this causes an answer that starts from the beginning of the sentence (i.e. `start_idx =0`) to be labeled as (0, 0). This might be related to #552 .
```
processor = SquadProcessor(...)
data_silo = DataSilo(processor=processor, batch_size=16, automatic_loading=False)
basic_texts = {"context": "endesa, s.a. financial statements for the year ended 31 december 2018 5 endesa, s.a. "
                          "and subsidiaries consolidated financial statements for the year ended 31 december 2018 207",
 "qas": [{"question": "What is the company name?", "id": "0",
          "answers": [{"text": "endesa", "answer_start": 0},
                      ], "is_impossible": False}]}

data_silo._load_data(train_dicts=[basic_texts])
```
```
print(data_silo.data['train'].datasets[0].tensors[6]) # labels
tensor([[[ 0,  0],
         [-1, -1],
         [-1, -1],
         [-1, -1],
         [-1, -1],
         [-1, -1]]])

print(data_silo.data['train'].datasets[0].tensors[-1]) # seq_2_start_t
tensor([8])
```

After changing the conditional statement to `passage_len > start_idx >= 0`:
```
print(data_silo.data['train'].datasets[0].tensors[6]) # labels
tensor([[[ 8, 10],
         [-1, -1],
         [-1, -1],
         [-1, -1],
         [-1, -1],
         [-1, -1]]])
print(data_silo.data['train'].datasets[0].tensors[-1]) # seq_2_start_t
tensor([8])
```

 - FARM version: 0.4.8

brandenchan:
Hey @EhsanM4t1qbit thanks for pointing this out! The new PR #564 implements the fix that you suggested. Can you confirm that this fixes your issue?

EhsanM4t1qbit:
Yes it does, thanks!
Do you think it makes sense to allow `passage_len` to be equal to  `end_idx` too? (i.e. `passage_len >= end_idx > 0`)

brandenchan:
@EhsanM4t1qbit Sorry I missed that last message!

[...]

Issue #552, opened 2020-09-22
Cannot extract QA answer at beginning of document


Timoeller:
**Describe the bug**
Answer at beginning of documents cannot be extracted.

**Error message**
No errors, just the predictions are wrong.

**Additional context**
I suspect the logit verification is not correct. We do check weather a start/end logit is within the question tokens, this is possibly flawed.
Doesnt predict correctly neither for bert nor for roberta type QA models.

**To Reproduce**
compare
```
inferencer = Inferencer.load(model_name_or_path="deepset/roberta-base-squad2", task_type="question_answering")

    qa_input = [
        {
            "qas": ["What is the largest city in Germany?"],
            "context": "Berlin is the capital and largest city of Germany by both area and population.",
        }
    ]

    results = inferencer.inference_from_dicts(qa_input)
    print(results)


vs 
```
    qa_input = [
        {
            "qas": ["What is the largest city in Germany?"],
            "context": "Document testing. With short text before Berlin it is still no answer. Berlin is the capital and largest city of Germany by both area and population.",
        }
    ]

vs
```
    qa_input = [
        {
            "qas": ["What is the largest city in Germany?"],
            "context": "Document testing this weird behaviour for all kinds of cases. With short text before Berlin it is still no answer. Berlin is the capital and largest city of Germany by both area and population.",
        }
    ]

ftesser:
I had the same feeling in some test I made, I noticed that a suspected percentage of un-predicted answers was located at the beginning of the documents.

antoniolanza1996:
Comment from https://github.com/deepset-ai/haystack/issues/476#issuecomment-708447639:

@Timoeller however, even though you have written that QA doesn't predict correctly neither for BERT nor for RoBERTa type QA models, I've noted that for BERT there are no problems. The problems come up **only with RoBERTa-Base and ELECTRA-Base**. Even with RoBERTa-Large and ELECTRA-Large is ok.

```python
from haystack.reader.farm import FARMReader
models = ["deepset/roberta-base-squad2","deepset/electra-base-squad2", "twmkn9/bert-base-uncased-squad2","twmkn9/albert-base-v2-squad2","deepset/minilm-uncased-squad2","ahotrod/electra_large_discriminator_squad2_512","phiyodr/roberta-large-finetuned-squad2"]
readers = []
for model in models:
  reader = FARMReader(model_name_or_path = model, use_gpu = True)
  readers.append(reader)

question = "What is the largest city in Germany?"
[...]

PR #564, opened 2020-10-05
Fix QA bug that rejected spans at beginning of passage

brandenchan:
Previously, the QA system could not predict spans that started at the first token of the passage. This PR fixes that and the related #558


Timoeller (approved):
K lets fix the label side and then move on to the model predictions.

@brandenchan when you work on the prediction side of answer beginning at start of text, please read the whole thread in issue #552

Issue #551, opened 2020-09-21
Natural Questions Inference return just one prediction result if there are more than one questions  


ftesser:
**Describe the bug**
Given a QA_input dictionary with more than one question e.g.:
```python
 QA_input = [
        {
            "qas": ["Did GameTrailers rated Twilight Princess as one of the best games ever created?", "Did GameTrailers rated Twilight Princess as one of the worst games ever created?"],
            "context":  "Twilight Princess was released to universal critical acclaim and commercial success. It received perfect scores from major publications such as 1UP.com, Computer and Video Games, Electronic Gaming Monthly, Game Informer, GamesRadar, and GameSpy. On the review aggregators GameRankings and Metacritic, Twilight Princess has average scores of 95% and 95 for the Wii version and scores of 95% and 96 for the GameCube version. GameTrailers in their review called it one of the greatest games ever created."
        }
    ]
```
the returned inference object:  
```python
result = model.inference_from_dicts(dicts=QA_input, return_json=False)
```
is a list that contains just one QApred object:
```python
[<farm.modeling.predictions.QAPred object at 0x7f50dcb73e20>]
```
that contains just the results for the first question.

**Expected behavior**
Do return a list of QAPred objects of the same size of the number questions listed in qas (like the "normal" Squad model does). 

**Additional context**
I understand that probably the reason of that is because the original NQ dataset contains just a question for each document, but I think that the API user it expect to receive answers for each question (or at least to be warned that Natural Questions Inference return just one prediction also of there are more questions).
   
**To Reproduce**
Modify  the `QA_input` in `examples/natural_questions.py` adding one or more questions (like above). 

Then run `examples/natural_questions.py` (to save time: the same error is obtained  running the script just from point 9 (https://github.com/deepset-ai/FARM/blob/c6871b67401eef36f9d4310c966063eda10b0dfe/examples/natural_questions.py#L125)

**System:**
 - OS: 
 - GPU/CPU:
 - FARM version:

Timoeller:
Hey thanks for raising this issue, agreed this is unexpected behaviour but I would not necessarily call it a bug. We are currently refactoring the input dicts and will possibly handle this case differently. 

As you suggested: a warning message would be nice for now. Would you like to create a PR with it?

ftesser:
Hello @Timoeller, yes I will make a PR for a warning message. 
But I really like to have this feature in FARM, I am going to explain why: 
[...]

PR #565, opened 2020-10-05
Added warning about that Natural Questions Inference. 

ftesser:
Work related to #551.

In this case we checking for both 'questions' and 'qas' present as key in the dictionary.


Timoeller (approved):
Looking good, thanks for including the check for qas as well.

Issue #569, opened 2020-10-06
Improve conversion from/to transformers


Timoeller:
- [x] Simplify usage with conversion class or similar
- [x] additional tests
- [x] Check conversion for tasks, e.g. see #570

Timoeller:
@bogdankostic is this fixed now with your PR #576 ? Please close this issue and the one for NER if it is fixed.

PR #576, opened 2020-10-08
Improve conversion from/to transformers

bogdankostic:
Related to #569


bogdankostic:
Conversion of multi-label text classification models is still not possible as there is no equivalent in transformers.
Also, conversion of roberta text classification models is still not possible. (As mentioned in the comments, we would need a separate RobertaClassificationHead for this.)

farm/modeling/adaptive_model.py
```diff
@@ -501,85 +503,16 @@ def get_language(self):
...
+        prediction head.
 
-        return transformers_model
+        :return: List of huggingface transformers models.
+        """
+        return Converter.convert_to_transformers(self)
 
     @classmethod
-    def convert_from_transformers(cls, model_name_or_path, device, task_type, processor=None):
+    def convert_from_transformers(cls, model_name_or_path, device, task_type=None, processor=None):
```
tholor, line 515: Why is task_type optional here?

farm/modeling/adaptive_model.py
```diff
@@ -813,3 +718,223 @@ def load_from_adaptive_model(cls, adaptive_model):
 
     def forward(self, *batch):
         return super().forward(input_ids=batch[0], padding_mask=batch[1], segment_ids=batch[2])
+
+
+class Converter:
```
tholor, line 723: How about we move this class to farm.conversion.transformers? adaptive_model.py is already huge

tholor (commented):
@bogdankostic Looking good. Only added a few comments. 
+ Can you please also update the example scripts examples/conversion_huggingface... ?

farm/modeling/adaptive_model.py
bogdankostic: Because we are inferring the task type automatically if task_type is not set. (https://github.com/deepset-ai/FARM/blob/aa39d7824a2e62d545a78df306a4684c407891f9/farm/modeling/adaptive_model.py#L800)

tholor: Oh ok, missed that :+1:

tholor (approved):
Looking good

PR #579, opened 2020-10-13
[RAG] Bumping up transformers version to 3.3.x

lalitpagaria:
Towards https://github.com/deepset-ai/haystack/issues/443

@tholor Please review

### DIFF
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -10,13 +10,19 @@ from typing import List
 
 import torch
 from torch import nn
-from torch.nn import CrossEntropyLoss, LayerNorm, MSELoss, BCEWithLogitsLoss
+from torch.nn import CrossEntropyLoss, MSELoss, BCEWithLogitsLoss
 
 from farm.data_handler.utils import is_json
 from farm.utils import convert_iob_to_simple_tags, try_get
 from farm.modeling.predictions import QACandidate, QAPred
 
 logger = logging.getLogger(__name__)
+
+try:
+    from apex.normalization.fused_layer_norm import FusedLayerNorm as BertLayerNorm
+except (ImportError, AttributeError) as e:
+    logger.info("Better speed can be achieved with apex installed from https://www.github.com/nvidia/apex .")
+    BertLayerNorm = torch.nn.LayerNorm
 
 
 class PredictionHead(nn.Module):
@@ -734,7 +740,7 @@ class BertLMHead(PredictionHead):
         # this is the "transform" module in the pytorch-transformers repo
         self.dense = nn.Linear(self.hidden_size, self.hidden_size)
         self.transform_act_fn = ACT2FN[self.hidden_act]
-        self.LayerNorm = LayerNorm(self.hidden_size, eps=1e-12)
+        self.LayerNorm = BertLayerNorm(self.hidden_size, eps=1e-12)
 
         # this is the "decoder" in the pytorch-transformers repo
         # The output weights are the same as the input embeddings, but there is
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -4,13 +4,13 @@ import os
 import numpy as np
 
 from pathlib import Path
-from transformers.modeling_bert import BertForPreTraining, BertLayerNorm, ACT2FN
+from transformers.modeling_bert import BertForPreTraining, ACT2FN
 from transformers.modeling_auto import AutoModelForQuestionAnswering, AutoModelForTokenClassification, AutoModelForSequenceClassification
 from typing import List
 
 import torch
 from torch import nn
-from torch.nn import CrossEntropyLoss, MSELoss, BCEWithLogitsLoss
+from torch.nn import CrossEntropyLoss, LayerNorm, MSELoss, BCEWithLogitsLoss
 
 from farm.data_handler.utils import is_json
 from farm.utils import convert_iob_to_simple_tags, try_get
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -17,7 +17,7 @@ sklearn
 seqeval
 mlflow==1.0.0
 # huggingface repository
-transformers==3.1.0
+transformers==3.3.1
```
lalitpagaria, line 20: I think better to have like this, but I am not sure about implication on FARM -
```
transformers>=3.3.1
```

lalitpagaria: I think it is bad idea to not use fixed version, after seeing these breaking changes :) 
https://github.com/huggingface/transformers/pull/1089/files

farm/modeling/prediction_head.py
```diff
@@ -734,7 +734,7 @@ def __init__(self, hidden_size, vocab_size, hidden_act="gelu", task_name="lm", *
         # this is the "transform" module in the pytorch-transformers repo
         self.dense = nn.Linear(self.hidden_size, self.hidden_size)
         self.transform_act_fn = ACT2FN[self.hidden_act]
-        self.LayerNorm = BertLayerNorm(self.hidden_size, eps=1e-12)
+        self.LayerNorm = LayerNorm(self.hidden_size, eps=1e-12)
```
lalitpagaria, line 737: Refer https://github.com/huggingface/transformers/pull/1089

lalitpagaria:
It seems model conversion test is failing but I am not able to pin point issue -
```
    def test_conversion_adaptive_model(caplog):
        if caplog:
            caplog.set_level(logging.CRITICAL)
    
        model = AdaptiveModel.convert_from_transformers("deepset/bert-base-cased-squad2", device="cpu", task_type="question_answering")
        transformer_model = model.convert_to_transformers()
        transformer_model2 = AutoModelForQuestionAnswering.from_pretrained("deepset/bert-base-cased-squad2")
        # compare weights
        for p1, p2 in zip(transformer_model.parameters(), transformer_model2.parameters()):
            if p1.shape != p2.shape:
                print("diff")
>           assert(p1.data.ne(p2.data).sum() == 0)
E           RuntimeError: The size of tensor a (768) must match the size of tensor b (2) at non-singleton dimension 0
```
These changes might give some clue https://github.com/huggingface/transformers/compare/v3.2.0...master (I don't suspect that changes in v3.3.0 or v3.3.1 causing this)

farm/modeling/prediction_head.py
```diff
@@ -18,6 +18,12 @@
 
 logger = logging.getLogger(__name__)
 
+try:
+    from apex.normalization.fused_layer_norm import FusedLayerNorm as BertLayerNorm
+except (ImportError, AttributeError) as e:
+    logger.info("Better speed can be achieved with apex installed from https://www.github.com/nvidia/apex .")
+    BertLayerNorm = torch.nn.LayerNorm
```
lalitpagaria, line 25: Refer this https://github.com/huggingface/transformers/pull/1089

requirements.txt
tholor: Yes, pinned version is better as there are still many breaking changes in transformers

farm/modeling/prediction_head.py
tholor: :+1:

tholor:
@bogdankostic Can you please have a look at this failing model conversion test? As you are currently refactoring the conversion (#576) we should make sure that it also works for transformers==3.3.1. So probably we should merge #576 first, then rebase or merge here and fix it if still failing.

bogdankostic:
I think I found the problem about the failing model conversion test. It seems to be related to this PR in transformers: https://github.com/huggingface/transformers/pull/7272
The consequence of this PR seems to be that in FARM, the models still contain pooling layers while in transformers, they don't. That seems to be the reason why the comparison of parameters fails. 

I'm not completely sure about the purpose of the pooling layer, but in their PR they describe that it is not needed and in fact not even used. How should we proceed?

bogdankostic:
Pooling layers are now removed for QA, NER and LM before converting FARM models to transformers in order to conform to the new transformers version.

tholor:
Great, I will run our QA performance benchmarks tomorrow. If they pass, we can merge.

Timoeller:
The QA accuracy benchmark shows some slight changes in performance. Especially the Evaluator drops by 0.7%:

Date | FARM commit | train EM | train f1 | train top5 | train elapsed | Evaluator EM | Evaluator f1 | Evaluator top1 | Evaluator elapsed | infer EM | infer f1 | infer elapsed
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
14.09.2020 | https://github.com/deepset-ai/FARM/commit/ce34cc24642e6c68edcb76275feac56ba71b0d34 | 0.780257 | 0.82286 | 0.974395 | 1123.208 | 0.777478 | 0.82156 | 0.84064 | 39 | 78.489 | 81.710402 | 26.472
16.10.2020 | https://github.com/deepset-ai/FARM/pull/579/commits/3ce5a97b3f8c9fe5b748d46dc479faa9f177258e | 0.78017 | 0.8234 | 0.97447 | 1120 | 0.770066 | 0.81432 | 0.83348 | 40 | 78.489 | 81.710402 | 26.48

tholor:
Hmm... not a big diff but still wonder where this comes from. What do you think @Timoeller ? Merging and later trying to understand the small diff if we find some time?

Timoeller:
I am fine with merging. We will deep dive into QA preds checking and other QA related stuff anyways soon.

How about merging this monday morning or do you depend on it now?

tholor:
Monday is fine

lalitpagaria:
Can we please merge this and create new Farm version. Not able to properly run transformers and haystack together even after many dirty hacks.

```
ERROR: farm 0.4.9 has requirement transformers==3.1.0, but you'll have transformers 3.3.1 which is incompatible.
```

PR #581, opened 2020-10-15
Fix import for Python3.6

tanaysoni:


Timoeller (approved):
quick fix for now, might move Converter somewhere else

PR #595, opened 2020-10-20
DPR transformers modelhub model load bugfix

kolk:
Bug fix for loading DPRQuestionEncoder and DPRContextEncoder weights from transformers modelhub

### DIFF
--- a/farm/modeling/language_model.py
+++ b/farm/modeling/language_model.py
@@ -1415,6 +1415,7 @@ class DPRQuestionEncoder(LanguageModel):
         super(DPRQuestionEncoder, self).__init__()
         self.model = None
         self.name = "dpr_question_encoder"
+        self.transformers_modelhub_path = "facebook/dpr-question_encoder-single-nq-base"
 
     @classmethod
     def load(cls, pretrained_model_name_or_path, language=None, **kwargs):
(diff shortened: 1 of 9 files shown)


farm/modeling/language_model.py
```diff
@@ -1404,8 +1405,13 @@ def load(cls, pretrained_model_name_or_path, language=None, **kwargs):
         else:
             # Pytorch-transformer Style
             dpr_question_encoder.model = transformers.DPRQuestionEncoder(config=transformers.DPRConfig(**kwargs))
-            # load weights from pretrained_model_name_or_path Language model into DPRQuestionEncoder
-            dpr_question_encoder.model.base_model.bert_model = AutoModel.from_pretrained(str(pretrained_model_name_or_path), **kwargs)
+            if pretrained_model_name_or_path == dpr_question_encoder.transformers_modelhub_path:
```
tholor, line 1408: Let's better check for if the string "dpr-question_encoder" is in pretrained_model_name_or_path. There might be more models / or local DPR models that we want to load directly...

Similarly for the context encoder below...

tholor (changes requested):

farm/modeling/language_model.py
kolk: Fixed it

tholor:
@kolk Can we close this one? I believe we tackle it in #606 now

PR #599, opened 2020-10-21
Fix random behavior when loading ELECTRA models

### LINKED ISSUE #519: Getting different predictions on different runs with same ELECTRA model.
I trained an Electra Model on text classification. 3 Classes.
Saved it with 

```
    save_dir = Path("saved_models/bert-german-doc-tutorial")
    model.save(save_dir)
    processor.save(save_dir)
```

Now I load Data I want to predict on and getting different results:

```
1st run:
labal_n          5055
label_i           855
label_e          8

2nd run
labal_n          4609
label_e          990
label_i           319

3rd run
label_i        3355
labal_n       2510
label_e       53
```

How on earth can that be? Even if the model would be a total crap the prediction should be deterministic - right?

bogdankostic:
This PR should fix #519. @brandenchan @stefan-it @PhilipMay 

### Summary of the bug
When training an ELECTRA model, saving it and loading it later on, the loaded model produced different (much worse) predictions than the trained model.

### Root of the bug
In this [transformers PR](https://github.com/huggingface/transformers/pull/4954), `summary_use_proj` was set by default to `True`, which means that a linear layer is stacked on top of pooling. Therefore, each time an ELECTRA model is loaded in FARM, the weights of this linear layer are randomly initialized.

### Fix
Set `summary_use_proj` to `False` in order to not add that linear layer.

### DIFF
--- a/farm/modeling/language_model.py
+++ b/farm/modeling/language_model.py
@@ -1275,6 +1275,7 @@ class Electra(LanguageModel):
         config.summary_last_dropout = 0
         config.summary_type = 'first'
         config.summary_activation = 'gelu'
+        config.summary_use_proj = False
         electra.pooler = SequenceSummary(config)
         electra.pooler.apply(electra.model._init_weights)
         return electra


tholor (approved):
Nice finding! This was a nasty one :)

Timoeller (approved):
Looking good. Thanks for the detailed description @bogdankostic 

What a big headache this behaviour created and what small changes needed to fix that headache.

PR #600, opened 2020-10-21
Infer model type from config

bogdankostic:
Until now, model types are inferred from the model name, which can cause problems like in #506. This PR makes it possible to infer the model type from the config files.

### DIFF
--- a/farm/modeling/language_model.py
+++ b/farm/modeling/language_model.py
@@ -42,7 +42,7 @@ from transformers.modeling_xlm_roberta import XLMRobertaModel, XLMRobertaConfig
 from transformers.modeling_distilbert import DistilBertModel, DistilBertConfig
 from transformers.modeling_electra import ElectraModel, ElectraConfig
 from transformers.modeling_camembert import CamembertModel, CamembertConfig
-from transformers.modeling_auto import AutoModel
+from transformers.modeling_auto import AutoModel, AutoConfig
 from transformers.modeling_utils import SequenceSummary
 from transformers.tokenization_bert import load_vocab
 import transformers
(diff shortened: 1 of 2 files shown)


farm/modeling/language_model.py
```diff
@@ -162,30 +162,68 @@ def load(cls, pretrained_model_name_or_path, n_added_tokens=0, language_model_cl
...
+            language_model_class = "Bert"
+        elif model_type == "xlnet":
+            language_model_class = "XLNet"
+        elif model_type == "electra":
+            language_model_class = "Electra"
+        elif model_type == "dpr":
+            if config.architectures[0] == "DPRQuestionEncoder":
+                language_model_class = "DPRQuestionEncoder"
+            elif config.architectures[0] == "DPRContextEncoder":
+                language_model_class = "DPRContextEncoder"
```
Timoeller, line 194: What abou dpr reader? 
Should we also fall back to string matching if the architecture doesnt match?

farm/modeling/tokenization.py
```diff
@@ -175,6 +141,97 @@ def load(cls, pretrained_model_name_or_path, tokenizer_class=None, use_fast=Fals
...
+            tokenizer_class = "BertTokenizer"
+        elif model_type == "xlnet":
+            tokenizer_class = "XLNetTokenizer"
+        elif model_type == "electra":
+            tokenizer_class = "ElectraTokenizer"
+        elif model_type == "dpr":
+            if config.architectures[0] == "DPRQuestionEncoder":
+                tokenizer_class = "DPRQuestionEncoderTokenizer"
+            elif config.architectures[0] == "DPRContextEncoder":
+                tokenizer_class = "DPRContextEncoderTokenizer"
```
Timoeller, line 181: dpr reader +
fall back to string matching here as well?

Timoeller (commented):
Looking already good.

I made a suggestion, though it might not possible to include dprReader for now (we are waiting for the transformers implementation)

farm/modeling/language_model.py
bogdankostic: DPR Reader is not supported at the moment (see https://github.com/deepset-ai/haystack/issues/440). Should we add a NotImplementedError?

Timoeller (approved):
Looking good

PR #605, opened 2020-10-27
Catch empty datasets in Inferencer

### LINKED ISSUE #454: QA inference breaks on large dataset
**Describe the bug**
QA inference failing on very large dataset

**Error message**
Inferencing Samples: 0 Batches [00:00, ? Batches/s]Traceback (most recent call last):
  File "question_answering_inference.py", line 89, in <module>
    question_answering()
  File "question_answering_inference.py", line 34, in question_answering
    result = model.inference_from_file(file=filename, return_json=False, multiprocessing_chunksize=1)
  File "/home/ubuntu/pycharm/timosyncNEW/farm/infer.py", line 645, in inference_from_file
    multiprocessing_chunksize=multiprocessing_chunksize, streaming=streaming)
  File "/home/ubuntu/pycharm/timosyncNEW/farm/infer.py", line 345, in inference_from_file
    streaming=streaming,
  File "/home/ubuntu/pycharm/timosyncNEW/farm/infer.py", line 637, in inference_from_dicts
    multiprocessing_chunksize=multiprocessing_chunksize, streaming=streaming)
  File "/home/ubuntu/pycharm/timosyncNEW/farm/infer.py", line 418, in inference_from_dicts
    return list(predictions)
  File "/home/ubuntu/pycharm/timosyncNEW/farm/infer.py", line 489, in _inference_with_multiprocessing
    dataset, tensor_names, baskets
  File "/home/ubuntu/pycharm/timosyncNEW/farm/infer.py", line 576, in _get_predictions_and_aggregate
    for i, batch in enumerate(tqdm(data_loader, desc=f"Inferencing Samples", unit=" Batches", disable=self.disable_tqdm)):
  File "/home/ubuntu/miniconda3/envs/py37/lib/python3.7/site-packages/tqdm/_tqdm.py", line 1060, in __iter__
    for obj in iterable:
  File "/home/ubuntu/miniconda3/envs/py37/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 345, in __next__
    data = self._next_data()
  File "/home/ubuntu/miniconda3/envs/py37/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 384, in _next_data
    index = self._next_index()  # may raise StopIteration
  File "/home/ubuntu/miniconda3/envs/py37/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 339, in _next_index
    return next(self._sampler_iter)  # may raise StopIteration
  File "/home/ubuntu/miniconda3/envs/py37/lib/python3.7/site-packages/torch/utils/data/sampler.py", line 200, in __iter__
    for idx in self.sampler:
  File "/home/ubuntu/miniconda3/envs/py37/lib/python3.7/site-packages/torch/utils/data/sampler.py", line 62, in __iter__
    return iter(range(len(self.data_source)))
TypeError: object of type 'NoneType' has no len()

**Additional context**
I suspect this has to do with one whole chunk in multiprocessing not being transformed correctly. We should remove empty datasets before sending them to _get_predictions_and_aggregate.

Timoeller:
fixes #454 
Only happens during multiprocessing.

When one multiprocessing chunk only contains invalid examples we run into errors. This PR catches these errors and produces an error message. It does not solve the root cause.

Invalid examples can be:
1. empty/malformed documents for classification,
2. empty/malformed documents for QA 
3. QA annotations where answer text and offsets do not conform

### DIFF
--- a/farm/infer.py
+++ b/farm/infer.py
@@ -505,21 +505,25 @@ class Inferencer:
         # Once a process spits out a preprocessed chunk. we feed this dataset directly to the model.
         # So we don't need to wait until all preprocessing has finished before getting first predictions.
         for dataset, tensor_names, baskets in results:
-            # TODO change format of formatted_preds in QA (list of dicts)
-            if aggregate_preds:
-                predictions = self._get_predictions_and_aggregate(
-                    dataset, tensor_names, baskets
-                )
+            if dataset is None:
+                logger.error(f"Part of the dataset could not be converted. Check previous log-messages for unconverted samples. \n"
+                             f"BE AWARE: The order of predictions should not conform with the input order!")
             else:
-                predictions = self._get_predictions(dataset, tensor_names, baskets)
+                # TODO change format of formatted_preds in QA (list of dicts)
+                if aggregate_preds:
+                    predictions = self._get_predictions_and_aggregate(
+                        dataset, tensor_names, baskets
+                    )
+                else:
+                    predictions = self._get_predictions(dataset, tensor_names, baskets)
 
-            if return_json:
-                # TODO this try catch should be removed when all tasks return prediction objects
-                try:
-                    predictions = [x.to_json() for x in predictions]
-                except AttributeError:
-                    pass
-            yield from predictions
+                if return_json:
+                    # TODO this try catch should be removed when all tasks return prediction objects
+                    try:
+                        predictions = [x.to_json() for x in predictions]
+                    except AttributeError:
+                        pass
+                yield from predictions
 
     @classmethod
     def _create_datasets_chunkwise(cls, chunk, processor):


tholor (commented):
This would fix the inferencer, but don't we have similar issues in other use cases (e.g. training https://github.com/deepset-ai/haystack/issues/488)? Maybe a similar fix inside DataSilo could tackle this?

Timoeller:
I thought so, too and used corrupted files for normal training (using the datasilo). I could not get QA processing to fail with these corrupted files.

I can try corrupting the files a bit more to see if we can reproduce datasilo errors. Will report back here.

Timoeller:
Ok I found out some strange behaviour when the context is either empty or does not contain the answer. I opened a separate issue and assigned @bogdankostic to it.

I would prefer to merge this solution for the inferencer now and work on the datasilo in a separate PR.

tholor (approved):
Ok sounds good. Let's tackle that separately

PR #607, opened 2020-10-28
Add option to set evaluation batch size

### LINKED ISSUE #604: Different batch size for training vs eval
Training needs much more memory than evaluation, since there are no gradients.

We need to have an optional eval batch size that should ideally be larger than training. Especially for large models with low training batch size eval becomes much slower (on the order of 5-10x).

bogdankostic:
Adds a separate batch size that is applied for dev set evaluation as well as test set evaluation.

Fixes #604.


Timoeller (approved):
Easy peasy.

PR #614, opened 2020-11-03
Ignore QA samples with empty context

### LINKED ISSUE #608: Error during QA dataprocessing when context is empty
**Describe the bug**
The problem is twofold:
1. If we have no text inside the document we get a Runtime error.
2. If we have text that does not contain the answer we do not have any warning messages.

**Error message**
```
  File "/foo/FARM/farm/data_handler/data_silo.py", line 176, in _get_dataset
    for dataset, tensor_names in results:
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 735, in next
    raise value
TypeError: 'NoneType' object is not iterable
```

**To Reproduce**
I uploaded a [sample file](https://gist.github.com/Timoeller/84da8f7bd41754a768f8c49896b365b8) to reproduce the error.
For problem 1. mentioned above take the file as is.
For problem 2. just insert any characters into the first context

bogdankostic:
Fixes #608 

With this PR we check if the context in QA samples is empty and if this is the case, we ignore that sample. Furthermore, we check if the answer string is contained in the context string. If this is not the case, we print out a warning.

### DIFF
--- a/farm/data_handler/processor.py
+++ b/farm/data_handler/processor.py
@@ -1240,6 +1240,14 @@ class SquadProcessor(QAProcessor):
 
         for index, document in zip(indices, dicts_tokenized):
             for q_idx, raw in enumerate(document):
+                # ignore samples with empty context
+                if raw["document_text"] == "":
+                    logger.warning("Ignoring sample with empty context.")
+                    continue
+                # check if answer string can be found in context
+                for answer in raw["answers"]:
+                    if answer["text"] not in raw["document_text"]:
+                        logger.warning(f"Answer '{answer['text']}' not contained in context.")
                 # In case of Question Answering the external ID is used for document IDs
                 id_external = try_get(ID_NAMES, raw)
                 id_internal = f"{index}-{q_idx}"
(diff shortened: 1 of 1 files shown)


Timoeller (changes requested):
As discussed, we need to check for more hidden errors further down the processing pipeline.

bogdankostic:
@Timoeller We didn't get a warning message with our example because we silently converted answers to no answers if the answer offsets exceeded the context length (see https://github.com/deepset-ai/FARM/blob/af6110175d3592165fd94891e99ef7a288f64e58/farm/data_handler/processor.py#L2195). I added a warning message.

Timoeller:
Transforming to no_answer makes sense. Thanks for adding the warning in that case.

Why did you delete the catching and removal of empty documents? I think we still want that simple case covered and later enhance the checks once we have erroneous documents as examples. Erroneous documents are docs filled with text but possibly malformed text or text unconvertible by the tokenizer (unsure if that is even possible : )

So please enable the removal of empty docs again.

bogdankostic:
The catching and removal of empty docs should still be there: https://github.com/deepset-ai/FARM/blob/af6110175d3592165fd94891e99ef7a288f64e58/farm/data_handler/processor.py#L1244

Timoeller:
Ah sorry - of course. I guess I was on the github tab that showed just the last changes from your latest commit. Late night coding... : )

Timoeller (approved):
Looking good.

PR #617, opened 2020-11-05
EarlyStopping fix and custom head

ftesser:
Hello, this PR, contains two commit related to EarlyStopping class:
 * the first fix a bug with the return in a particular case
 * the second is an improvement that allow the user to choose the prediction head to refer to for the metric (useful for Natural Questions).


Timoeller (approved):
Looking good.

Thanks for creating the new PR with only relevant changes. 
Nice bug fix catch there btw and you also added a docstring for the head. I wish all external PRs would be like that : )

PR #619, opened 2020-11-09
Add MultiGPU support for DPR Training via DDP

tholor:
In order to enable larger batch sizes for DPR training, we need multi GPU support. 
Let's use DistributedDataParallel as it's the more performing and scalable option ...

- [x] gather tensors for loss with in-batch negatives
- [x] verify eval is only running on rank 0 
- [x] adjust vocab size check for DDP
- [x] verify distribution of dataset into batches
- [x] infer/pass distributed_world_size in prediction head
- [x] fix nonzero() deprecation warning

Future work
- [ ] refactor `all_gather_list` to torch's standard `all_gather()`

### DIFF
--- a/examples/dpr_encoder.py
+++ b/examples/dpr_encoder.py
@@ -3,6 +3,8 @@ import logging
 import os
 import pprint
 from pathlib import Path
+import argparse
+
 
 from farm.data_handler.data_silo import DataSilo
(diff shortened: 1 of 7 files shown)


Timoeller (commented):
Ok lets merge this PR now and improve it later on.

PR #624, opened 2020-11-13
Increase transformers version to 3.5.1

Timoeller:


test/test_tokenization.py
```diff
@@ -272,8 +271,8 @@ def test_fast_bert_tokenizer_strip_accents(caplog):
                                use_fast=True,
                                strip_accents=False)
     assert type(tokenizer) is BertTokenizerFast
+    assert tokenizer.do_lower_case
     assert tokenizer._tokenizer._parameters['strip_accents'] is False
```
Timoeller, line 275: Hey @PhilipMay I am confused about the usage of `strip_accents` and how it is used in the fast tokenizers. Since you added it to transformers could you have a look at this failing test?

PhilipMay: Sure. :-) 

This is the problem right?
```
caplog = <_pytest.logging.LogCaptureFixture object at 0x7fa228117a50>

    def test_fast_bert_tokenizer_strip_accents(caplog):
        caplog.set_level(logging.CRITICAL)
    
        tokenizer = Tokenizer.load("dbmdz/bert-base-german-uncased",
                                   use_fast=True,
                                   strip_accents=False)
        assert type(tokenizer) is BertTokenizerFast
        assert tokenizer.do_lower_case
>       assert tokenizer._tokenizer._parameters['strip_accents'] is False
E       AttributeError: 'tokenizers.Tokenizer' object has no attribute '_parameters'
```

The old tokenizer implementation was like this:

- if ippercase then do not strip accents
- if lowercase then always strip accents

I wanted to bo able to do lower case but not to strip accents. Do I introduced a new param called `strip_accents` - see #6280

`strip_accents` has the default value of `None` which results in the "old" behavior. This is because I did not want to introduce a breaking change.

See here:
```
            **strip_accents**: (`optional`) boolean (default None)
                Whether to strip all accents. If this option is not specified (ie == None),
                then it will be determined by the value for `lowercase` (as in the original Bert).
```

If `strip_accents` is set to True or False it works like specified by the default value.
Because `dbmdz/bert-base-german-uncased` is an old model and does not specify `strip_accents` in its tokenizer config file thing the value should be at default (`None`).

So you should change it it `assert tokenizer._tokenizer._parameters['strip_accents'] is None`

And introduce a 2nd test where you use our electra model `german-nlp-group/electra-base-german-uncased` and test for `assert tokenizer._tokenizer._parameters['strip_accents'] is False`.

Timoeller: Ah ok, I understand the parameter better now. thanks.

But the error is coming from `'tokenizers.Tokenizer' object has no attribute '_parameters'`. It seems to me as something has changed fundamentally in the tokenizer classes. Do you agree?

PhilipMay: I do not think so. The FastTokenizer is just implemented in RUST and you can not access the internal properties with Python code.

AFAIK it is set here: https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_bert_fast.py#L199

PhilipMay: I do not think so. The FastTokenizer is just implemented in RUST and you can not access the internal properties with Python code.

AFAIK it is set here: https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_bert_fast.py#L199

But `type(pre_tok_class)` is `tokenizers.normalizers.BertNormalizer` which is not implemented with Python.
See here: https://github.com/huggingface/tokenizers/blob/master/bindings/python/py_src/tokenizers/normalizers/__init__.py#L4

And here: https://github.com/huggingface/tokenizers/blob/fff856cff76c1cb86d3f1ec392b5313f93b76ce0/bindings/python/py_src/tokenizers/normalizers/__init__.pyi#L18

Timoeller: Ok, does this mean I can delete the test?

To confirm, the test was passing before on the `BertTokenizerFast` and somehow accessed the "_parameters" attribute from within python. So you are saying that a lot of the parts in the BertTokenizerFast that were previously in python now got ported to the rust part and are not accessible?

PhilipMay: That test was green with transformers version 3.4.0 right? And you are 100% sure that it did load the fast tokenizer? Or could it be that the slow tokenizer somehow sneaked in?

Timoeller: Exactly, that test is green right now and there is an assertion that the fast tokenizer is used, so I guess it loads it as well.

bogdankostic:
My latest commits should make the use of Fast Tokenizers possible, therefore this fixes #634. I was not sure about the default value for `use_fast`. As I don’t see any problem with using Fast Tokenizers, I set it to `True`. Feel free to reset it to `False` if there is still something that blocks the use of Fast Tokenizers as default.

My latest commits should also fix #516, so PR #518 can be closed.

farm/data_handler/input_features.py
```diff
@@ -416,40 +423,25 @@ def sample_to_features_qa(sample, tokenizer, max_seq_len, sp_toks_start, sp_toks
...
+        # Fast Tokenizer does not support encode_plus with pretokenized input, so we just concatenate the input ids
+        question_input_ids = tokenizer.convert_tokens_to_ids(sample.tokenized["question_tokens"])
+        passage_input_ids = tokenizer.convert_tokens_to_ids(sample.tokenized["passage_tokens"])
+
+        input_ids = tokenizer.build_inputs_with_special_tokens(token_ids_0=question_input_ids,
+                                                               token_ids_1=passage_input_ids)
+        segment_ids = tokenizer.create_token_type_ids_from_sequences(token_ids_0=question_input_ids,
+                                                                     token_ids_1=passage_input_ids)
+
+        len_mid_sep_tokens = len(input_ids) - len(question_input_ids) - len(passage_input_ids) - 2
```
tholor, line 442: Why -2? Is that really generic for all model types?

farm/data_handler/processor.py
```diff
@@ -1290,6 +1709,44 @@ def _sample_to_features(self, sample) -> dict:
                                          max_answers=self.max_answers)
         return features
 
+    def _dicts_to_baskets_samples_and_features(self, dicts, indices) -> [SampleBasket]:
+        """This method is used so that we need to tokenize only once when using a fast tokenizer."""
+        # Perform tokenization on documents and questions resulting in an unnested list of doc-question pairs
+        dicts_tokenized = [_apply_tokenization(d, self.tokenizer) for d in dicts]
```
tholor, line 1715: @brandenchan can you please review this part for QA?

farm/data_handler/processor.py
```diff
@@ -1596,6 +2052,37 @@ def _sample_to_features(self, sample: Sample) -> dict:
                                          max_answers=self.max_answers)
         return features
 
+    def _dict_to_samples_and_features(self, dictionary: dict, **kwargs) -> [Sample]:
```
tholor, line 2055: @brandenchan ... and similarly for NQ

farm/data_handler/processor.py
```diff
@@ -334,6 +338,9 @@ def _dict_to_samples(cls, dictionary: dict, all_dicts=None) -> [Sample]:
     def _sample_to_features(cls, sample: Sample) -> dict:
         raise NotImplementedError()
 
+    def _dict_to_samples_and_features(self, dictionary: dict, all_dicts=None) -> [Sample]:
```
tholor, line 341: Why do you want to introduce a new method `_dict_to_samples_and_features`? 
Wouldn't it be cleaner to:
-  change `dict_to_samples` to also create the features 
- deprecate `_sample_to_features`
- eventually get rid of "tokenize_with_metadata"? 

In that case we would not have so much parallel code (once for slow, once for fast tokenizer) and it would be easier to maintain. Do you see any problems running the same code with the slow tokenizer?

tholor (commented):
Great progress! Left a few comments. 
Biggest point: Do we really want to introduce a new `_dict_to_samples_and_features()`? We should have a design discussion here as I would rather prefer refactoring the existing methods and avoid duplicate code.

Timoeller:
Changes made here have been merged with #649 
Closing now

Issue #610, opened 2020-10-29
Percentage of masked tokens in LM task


pashok3d:
**Question**
How can I define the probability of masking each token for LM task?

**Additional context**
As far as I know, the probability of masking each token for the LM task is 0.15. Is there a way to change it?

Timoeller:
Hey thanks for using FARM. This feature is actually annoyingly hidden from the user.

You can set the masked_lm_prob parameter inside the mask_random_words() function used [here in FARM](https://github.com/deepset-ai/FARM/blob/78fb5cfb02cdb38f01b66d67f9ab640e6da4e79b/farm/data_handler/input_features.py#L266) or depending on your finetuning style a little bit below the linked lined.

If you want to extent FARM by making this an optional Parameter of the BertStyleLMProcessor Class we are more than happy to accept a PR.

Timoeller:
Could you resolve your issue with the parameter I showed you? If so please close the issue (or even create a PR with your solution. Would be happy to include it if it is simple).

pashok3d:
#630

Timoeller:
Thanks for implementing the functionality. Great PR that I could just merge as is!

PR #630, opened 2020-11-16
Adding probability of masking a token parameter for LM task 

pashok3d:
#610


Timoeller (approved):
Nice one! Functionality added without breaking or altering existing code.

PR #631, opened 2020-11-16
Conversion from tf to pytorch: Remove duplicate, fix import

Timoeller:
FYI @tanaysoni I cleaned up the conversion code that did not make sense to me (unsure how you got it working in the first place anyways :)


mchari:
Apologies, I should have pointed out one more change that I had to make to progress ...
load_tf_weights_in_bert(model, config, tf_checkpoint_path) <--- needed to add config too, at least for tf 2.3.0

Timoeller:
k makes sense. It is a non default parameter in the function and needs to be set.
Fixed in https://github.com/deepset-ai/FARM/commit/f1c02ee99d090ce7e986d0fe1c4598af592287f9

Issue #611, opened 2020-10-29
Inference Benchmarks of QA models are off 


tholor:
**Describe the bug**
EM and F1 are significantly low when running the [reader benchmarks](https://github.com/deepset-ai/haystack/blob/add_cml_benchmarks/test/benchmarks/reader.py) in Haystack.

Latest Master
|    |       EM |      f1 |   top_n_accuracy |   top_n |   reader_time |   seconds_per_query |   passages_per_second | reader                         | error   |
|---:|---------:|--------:|-----------------:|--------:|--------------:|--------------------:|----------------------:|:-------------------------------|:--------|
|  0 | 0.323024 | 0.37967 |          0.95584 |       5 |       123.854 |           0.0104377 |               99.7142 | deepset/bert-base-cased-squad2 |         |

FARM 0.4.9
|    |       EM |       f1 |   top_n_accuracy |   top_n |   reader_time |   seconds_per_query |   passages_per_second | reader                      | error   |
|---:|---------:|---------:|-----------------:|--------:|--------------:|--------------------:|----------------------:|:----------------------------|:--------|
|  0 | 0.758975 | 0.806799 |         0.967133 |       5 |       135.263 |           0.0113992 |               91.3039 | deepset/roberta-base-squad2 |         |

However, it doesn't seem to be any of the latest changes as already with this commit I get bad results.

FARM @ dac388aa00eb63ad12452f97e40f51a218110b90 (Oct 21) 
|    |       EM |       f1 |   top_n_accuracy |   top_n |   reader_time |   seconds_per_query |   passages_per_second | reader                      | error   |
|---:|---------:|---------:|-----------------:|--------:|--------------:|--------------------:|----------------------:|:----------------------------|:--------|
|  0 | 0.330777 | 0.383186 |         0.963341 |       5 |       134.598 |           0.0113432 |               91.7545 | deepset/roberta-base-squad2 |         |

**Expected behavior**
Similar EM and F1

**Additional context**
I tried to switch transformers from 3.3.1 back to 3.1.0 (as it was used in FARM 0.4.9) => no effect

**To Reproduce**
Run the above benchmark script from the haystack branch

**System:**
 - OS: Ubuntu 18.04 
 - GPU/CPU: GPU
 - FARM version: latest master  / 0.4.9

tholor:
Similar for 2ab42d2

brandenchan:
Results are high again at c106bca (Oct 4)

brandenchan:
[...]

PR #635, opened 2020-11-18
Remove empty string span answers

brandenchan:
If the input raw_dict["answers"] contains an answer where text="" we should remove it. Proper no_answer format is to have raw_dict["answers"]=[].

Addresses #611


tholor (approved):
LGTM

Issue #636, opened 2020-11-18
Tests fail on Windows


ftesser:
**Describe the bug**
After installing farm and its dependencies on Windows, running the tests produce the error below.
This does not happen on Linux OS.

**Error message**
```
ImportError while loading conftest 'C:\builds\fairmat\dev\FARM\test\conftest.py'.
conftest.py:7: in <module>
    from farm.modeling.adaptive_model import AdaptiveModel
..\venv\lib\site-packages\farm\modeling\adaptive_model.py:17: in <module>
    from farm.modeling.prediction_head import PredictionHead, pick_single_fn
..\venv\lib\site-packages\farm\modeling\prediction_head.py:14: in <module>
    from torch.distributed import all_gather
E   ImportError: cannot import name 'all_gather' from 'torch.distributed' (c:\builds\fairmat\dev\farm\venv\lib\site-packages\torch\distributed\__init__.py)
```

**Expected behavior**
Successfully run the tests also on Windows.

**Additional context**
- pytorch version: 1.6.0

Looking to `torch.distributed.__init__`  I have noticed that `all_gather` probably is imported only for Linux and MacOS:   

```python
def is_available():
    """
    Returns ``True`` if the distributed package is available. Otherwise,
    ``torch.distributed`` does not expose any other APIs. Currently,
    ``torch.distributed`` is available on Linux and MacOS. Set
    ``USE_DISTRIBUTED=1`` to enable it when building PyTorch from source.
    Currently, the default value is ``USE_DISTRIBUTED=1`` for Linux and
    ``USE_DISTRIBUTED=0`` for MacOS.
    """
    return hasattr(torch._C, "_c10d_init")
```

**To Reproduce**

```bash
cd test && pytest
```
**System:**
 - OS: Windows
 - GPU/CPU: CPU
 - FARM version: 0.5.0

Timoeller:
This would normally be quite difficult to debug, but I do not find any usage of all_gather in our code currently.

@tholor could you verify this, since you added this import during the MultiGPU DDP support for DPR and if it is needed.

ftesser:
Thanks @Timoeller, in the meantime I can run the tests without the import line `from torch.distributed import all_gather` in `farm\modeling\prediction_head.py`.

Timoeller:
I added a potential solution to #637

Timoeller:
#637 merged.

Will also update to newer pytorch soon, which should reduce errors related to DDP as well.

PR #637, opened 2020-11-19
Make tests working on Windows

ftesser:
This PR is related to #636.
It will be an incremental PR, since I see that after this first commit, some others changes are needed to successfully run all the tests on Windows.

This first commit remove  the not used 'from torch.distributed import all_gather' from 'farm\modeling\prediction_head.py'. There is also a newline added here.

### DIFF
--- a/farm/data_handler/utils.py
+++ b/farm/data_handler/utils.py
@@ -116,7 +116,7 @@ def read_jsonl(file, proxies=None):
     if not (os.path.exists(file)):
         logger.info(f" Couldn't find {file} locally. Trying to download ...")
         _download_extract_downstream_data(file, proxies=proxies)
-    dicts = [json.loads(l) for l in open(file)]
+    dicts = [json.loads(l) for l in open(file, encoding="utf-8")]
     return dicts
 
 def read_ner_file(filename, sep="\t", proxies=None):
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -11,7 +11,6 @@ from typing import List, Tuple
 import torch
 from torch import nn
 from torch.nn import CrossEntropyLoss, MSELoss, BCEWithLogitsLoss, NLLLoss
-from torch.distributed import all_gather
 from farm.data_handler.utils import is_json
 from farm.utils import convert_iob_to_simple_tags, try_get, all_gather_list
 from farm.modeling.predictions import QACandidate, QAPred
--- a/test/test_dpr.py
+++ b/test/test_dpr.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 import torch
 import logging
@@ -10,6 +11,7 @@ from farm.modeling.tokenization import Tokenizer
 from farm.utils import set_all_seeds, initialize_device_settings
 from farm.data_handler.dataset import convert_features_to_dataset
 
+@pytest.mark.skipif(os.name == 'nt', reason="DDP does not work on Windows")
 def test_dpr_modules(caplog=None):
     if caplog:
         caplog.set_level(logging.CRITICAL)


ftesser:
After the first commit, I noticed this error on Windows test (note that from side the test on Linux always run successfully):

```
test_natural_questions.py:43: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\venv\lib\site-packages\farm\data_handler\data_silo.py:113: in __init__
    self._load_data()
..\venv\lib\site-packages\farm\data_handler\data_silo.py:215: in _load_data
    self.data["train"], self.tensor_names = self._get_dataset(train_file)
..\venv\lib\site-packages\farm\data_handler\data_silo.py:141: in _get_dataset
    dicts = list(self.processor.file_to_dicts(filename))
..\venv\lib\site-packages\farm\data_handler\processor.py:1383: in file_to_dicts
    dicts = read_jsonl(file, proxies=self.proxies)
..\venv\lib\site-packages\farm\data_handler\utils.py:119: in read_jsonl
    dicts = [json.loads(l) for l in open(file)]
..\venv\lib\site-packages\farm\data_handler\utils.py:119: in <listcomp>
    dicts = [json.loads(l) for l in open(file)]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <encodings.cp1252.IncrementalDecoder object at 0x00000142D4728280>
input = b'edia.org/w/index.php?title=List_of_WWE_Champions&oldid=866450897 \'\' Categories : <Ul> <Li> World heavyweight wrest...evel": false, "end_token": 5090}, {"start_token": 5090, "top_level": false, "end_token": 5204}, {"start_token": 5204, '
final = False

    def decode(self, input, final=False):
>       return codecs.charmap_decode(input,self.errors,decoding_table)[0]
E       UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 1012: character maps to <undefined>

c:\python\lib\encodings\cp1252.py:23: UnicodeDecodeError
```

So the second commit is need to fix that.

ftesser:
After the first two commits, there is only one test that fail on Windows:  test.test_dpr.test_dpr_modules
The reason is: 
```
E           AttributeError: module 'torch.distributed' has no attribute 'get_rank'

..\venv\lib\site-packages\farm\modeling\prediction_head.py:1644: AttributeError
```
I see that this refer to `logits_to_loss` method on  prediction_head.py :


```python
# Check if DDP is initialized
        try:
            rank = torch.distributed.get_rank()
        except AssertionError:
            rank = -1
```

Any suggestions to make this working also on windows test? 
If DDP does not work on windows perhaps we can exclude this test when running on Windows?

ftesser:
My current proposal is to exclude `test.test_dpr.test_dpr_modules` just when pytest is run on a Windows machine.
If you think that it is possible somehow to test `dpr_modules` without `DDP` it would probably be better, but for the moment this solution allows you to test almost everything even on windows.

Let me know what do you think.

Timoeller:
Hey @ftesser thanks for looking into this.
Generally I believe it doesnt make much sense having different code/tests for different distributions. It will be tedious to keep track of all the individual changes.

The DDP module is quite important for making DPR trainable so excluding it would not be preferred.

Looking into Pytorch support for DDP in windows I found this PR: https://github.com/pytorch/pytorch/pull/45335 which was merged on Sep 25th. So this code is only in the newest pytorch 1.7.0 release (we are currently using pytorch 1.6.0).

So actionable insights: We want to do a FARM release later this week, after this we can increment the pytorch version and see if this fixes the DDP issues. Would that be a solution for you? Of course you can try updating pytorch beforehand yourself, we would appreciate your insights.

ftesser:
Thanks @Timoeller for your feedback.
Of course if with pytorch 1.7.0 DDP will works on Windows this is the best solution, so I am ok to wait for that. I will try to updating pytorch beforehand myself.

Just a note about this PR, commit https://github.com/deepset-ai/FARM/pull/637/commits/0aedd051e4f80dcefd9948cfba7e31ae37e0384c solve a UnicodeDecodeError, this should be applied in all cases.

Timoeller:
Sorry for the delay @ftesser
We actually made huge changes in #649 that caused a lot of testing.

We also quickly tested updating to pytorch in #660 which produced failing onnx conversion tests. Unfortunately we will be able to look into this with the start of the new year.

I will merge you PR now to include this nice patch in the coming release.

PR #638, opened 2020-11-24
Remove quadratic loop in _init_samples_in_baskets for efficiency

brandenchan:
There is a loop within a loop in Processor._init_samples_in_baskets(self) that causes computation to scale quadratically with the number of samples. This should resolve speed issues mentioned [here](https://github.com/deepset-ai/haystack/issues/602).

### DIFF
--- a/farm/data_handler/processor.py
+++ b/farm/data_handler/processor.py
@@ -335,8 +335,8 @@ class Processor(ABC):
         raise NotImplementedError()
 
     def _init_samples_in_baskets(self):
+        all_dicts = [b.raw for b in self.baskets]
         for basket in self.baskets:
-            all_dicts = [b.raw for b in self.baskets]
             try:
                 basket.samples = self._dict_to_samples(dictionary=basket.raw, all_dicts=all_dicts)
                 for num, sample in enumerate(basket.samples):


brandenchan:
@tholor @tanaysoni This change only makes sense as long as _dict_to_samples() doesn't change all_dicts in any way. I assume that the line I moved was inserted in the first place for the sake of pretraining where we need a copy of all dicts to randomly pick a next sentence. Can either of you confirm this?

brandenchan:
Currently running tests to see whether all_dicts changes during preprocessing. 
- [ ] Squad
- [ ] Natural Questions 
- [ ] LM Pretraining

tholor (approved):
LGTM

Timoeller:
Ok, nice analysis! Can we remove the unnecessary param all_dicts for NQ then (the NQ processing is complex already and simplifying redundant parts would be great)?

tanaysoni (approved):

brandenchan:
@Timoeller `all_dicts` is passed in to NaturalQuestionsProcessor._dict_to_samples()` because it is part of the signature of `Processor._dict_to_samples()`. It is used in `BertStyleLMProcessor._dict_to_samples()`. Because of this I'm currently in favor of leaving it as it is.

Timoeller:
Convinced @brandenchan . Removing it just for NQ doesnt make any sense at this processing step. LGTM

Issue #590, opened 2020-10-16
Improve Error Messages / Logs


Timoeller:
We can improve Error messages to be more informative. This is a non comprehensive list, feel free to add incomprehensible error messages in the issue thread:

Prio 1
- [x] Only show for real download: "Could not find deepset/roberta-base-squad2 locally. Try to download from model hub ..."
- [x] Remove: "WARNING - farm.modeling.language_model - Could not automatically detect from language model name what language]it is. ..."
- [x]  Error if no labels are produced by processor at training time (probably together with a ascii overview of connections between PH and tasks)
- [x]  when passing in an empty list as label list it breaks. passing anything (not matching the labels or even the number of different labels present in data) makes it work. This might brake things later on when evaluating (Run NER)
- [x]  passing the wrong train filename produces a cryptic error message: ValueError: Unexpected character found when decoding 'true' (Try QA)
- [x] for QA: do not display all cases where answer text is not euqals answer position. Print out stats how many examples could not be converted and maybe the ids
- [x] deprecate message for inference_from_objects (when we call inference from haystack) (don't call inference_from_objects in haystack)
- [x] Move unimportant messages from info to debug level
- [x] Reduce logged samples from 2 per dataloader (train,dev,test) to 1x for train and 1x for dev

Prio 2: 
- [x]  "Could not convert labels to ids via label_list!
- [ ]  If your are running in *inference* mode: Don't worry!" is displayed for every sample in NER inference mode. Lets find a switch to display it only once.

tholor:
- [ ] Only show for real download: "Could not find `deepset/roberta-base-squad2` locally. Try to download from model hub ..."
- [ ] Remove: 11/04/2020 06:44:26 - WARNING - farm.modeling.language_model -   Could not automatically detect from language model name what language it is. ..."

PR #639, opened 2020-11-26
Clean logging and error handling

brandenchan:
This PR addresses many of the issues in #590 regarding error handling and logging.

### DIFF
--- a/farm/data_handler/data_silo.py
+++ b/farm/data_handler/data_silo.py
@@ -129,7 +129,7 @@ class DataSilo:
         """
         dicts = [d[1] for d in chunk]
         indices = [x[0] for x in chunk]
-        dataset = processor.dataset_from_dicts(dicts=dicts, indices=indices)
+        dataset = processor.dataset_from_dicts(dicts=dicts, indices=indices, return_problematic=True)
         return dataset
 
     def _get_dataset(self, filename, dicts=None):
@@ -181,10 +181,12 @@ class DataSilo:
             if filename:
                 desc += f" {filename}"
             with tqdm(total=len(dicts), unit=' Dicts', desc=desc) as pbar:
-                for dataset, tensor_names in results:
+                for dataset, tensor_names, problematic_samples in results:
                     datasets.append(dataset)
                     # update progress bar (last step can have less dicts than actual chunk_size)
                     pbar.update(min(multiprocessing_chunk_size, pbar.total-pbar.n))
+                    self.processor.problematic_sample_ids.update(problematic_samples)
+            self.processor.log_problematic()
             # _dataset_from_chunk can return a None in cases where downsampling has occurred
             datasets = [d for d in datasets if d]
             concat_datasets = ConcatDataset(datasets)
@@ -201,6 +203,7 @@ class DataSilo:
         :param test_dicts: (Optional) dicts containing examples for test.
         :return: None
         """
+
         logger.info("\nLoading data into the data silo ..."
                     "{}".format(TRACTOR_SMALL))
         # train data
@@ -218,6 +221,7 @@ class DataSilo:
         else:
             logger.info("No train set is being loaded")
             self.data["train"] = None
+        self.processor.log_problematic()
 
         # dev data
         logger.info("")
@@ -239,6 +243,7 @@ class DataSilo:
         else:
             logger.info("No dev set is being loaded")
             self.data["dev"] = None
+        self.processor.log_problematic()
 
         logger.info("")
         logger.info("LOADING TEST DATA")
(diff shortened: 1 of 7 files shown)


brandenchan:
As it is right now, I have removed sample level error messages in NER that would, for example, say that a given sample has a label not in the label list, or that it doesn't have a label field at all. 

Instead, failed featurization is handled at a higher level so that we get a single line mentioning how many samples are not properly featurized. However, since we chunk our data, we actually have **one line per chunk** that states how many problematic samples there are.

Pros:
- Reduced number of error printouts

Cons:
- No sample specific error messages
- Potentially still quite a lot of lines of error logging

If we really want a single line of error printout, we will need to make changes to some very high level fns that will have impact across all tasks (e.g. `Datasilo._dataset_from_chunk( )` or `Processor.dataset_from_dicts( )` )

@Timoeller  thoughts?

farm/data_handler/processor.py
```diff
@@ -1227,15 +1237,21 @@ def dataset_from_dicts(self, dicts, indices=None, return_baskets=False):
         self._init_samples_in_baskets()
         self._featurize_samples()
         if 0 in indices:
-            self._log_samples(2)
+            self._log_samples(1)
         # This mode is for inference where we need to keep baskets
         if return_baskets:
             dataset, tensor_names = self._create_dataset(keep_baskets=True)
-            return dataset, tensor_names, self.baskets
+            ret = [dataset, tensor_names]
```
Timoeller, line 1244: I am not sure about the consequences, but lists are mutable and tuples (the return type before) are not. This might introduce some pretty weird behaviour. Do we need it?

Please also check the other return statement above

farm/data_handler/data_silo.py
```diff
@@ -181,10 +181,12 @@ def _get_dataset(self, filename, dicts=None):
             if filename:
                 desc += f" {filename}"
             with tqdm(total=len(dicts), unit=' Dicts', desc=desc) as pbar:
-                for dataset, tensor_names in results:
+                for dataset, tensor_names, problematic_samples in results:
```
Timoeller, line 184: Doesnt this statement break in case when the problematic samples are not returned?

farm/data_handler/processor.py
```diff
@@ -346,14 +347,23 @@ def _init_samples_in_baskets(self):
...
             for sample in basket.samples:
                 try:
                     sample.features = self._sample_to_features(sample=sample)
                 except Exception as e:
-                    logger.error(f"Could not convert this sample to features: \n {sample}")
-                    logger.error(f"Basket id: id_internal: {basket.id_internal}, id_external: {basket.id_external}")
-                    logger.error(f"Error message: {e}")
+                    curr_problematic_sample_ids.append(sample.id)
+        if curr_problematic_sample_ids:
+            self.problematic_sample_ids.update(curr_problematic_sample_ids)
```
Timoeller, line 358: Why are we updating the problematic_sample_ids here and when we collect results from MP?

Timoeller (changes requested):
I think with problematic_ids we need some further testing, especially if other tasks break.

farm/data_handler/processor.py
brandenchan: I wrote it like this to avoid having too many nested if statements since both `return_baskets` and `return_problematic` can alter the number of objects that are returned. Converting `ret` into a tuple in the next commit to avoid this mutability.

farm/data_handler/data_silo.py
brandenchan: In this context, `results` is always the output of `_dataset_from_chunk( )` which I modified to always return 3 objects so I don't think this unpacking is an issue. As an asside, this function (i.e. `_get_dataset( )`) is the only method in our code base which calls `_dataset_from_chunk( )` so it shouldn't be problematic that I've modified it to return 3 objects.

farm/data_handler/processor.py
brandenchan: The idea here is that we have two levels at which `problematic_ids` are stored. Within multiprocessing, each Processor needs to collect `problematic_ids` as it iterates through the baskets in its chunks. These are eventually returned out of multiprocessing via `dataset_from_dicts( )`. At this point, these returned `problematic_ids` are [collected together in the non-multiprocessing Processor ](https://github.com/deepset-ai/FARM/blob/ca358df76bd7e6ec952d22eb9b90e3c7ae733c7a/farm/data_handler/data_silo.py#L188). 

I went for this approach because it allows for us to implement logging problematic samples neatly as a method of the processor (`Processor.log_problematic( )`) as long as the problematic sample ids are stored in the `Processor`. 

I do admit that it might be a little confusing coordinating between the higher level single `Processor` and the multiple lower level `Processor`s. Do you have any thoughts or suggestions on this?

farm/data_handler/data_silo.py
Timoeller: Ok, agreed, but this means we always need to return those ids. 
We do have a flag in dataset_from_dicts, return_problematic=True, which seems to break the processing when set to false, right?

Issue #628, opened 2020-11-16
Add support for model versioning in HF model hub


brandenchan:
Transformers 3.5 now allows for different versions of models to be uploaded. This introduces a new argument "revision" when loading models. We should add support for this argument in FARM so that we can download specific model versions.

We should wait for this #624 to be completed first

PR #648, opened 2020-12-03
Support Model Hub's model versioning

brandenchan:
Addresses #628. Allows for the new revision argument that defines which version of a model from HuggingFace's Model Hub to use. 

This PR adds the argument to the `load( )` functions of the following classes:
- Inferencer
- Tokenizer
- LanguageModel
- PredictionHead
- Converter

This should probably not be merged before #624

### DIFF
--- a/farm/conversion/transformers.py
+++ b/farm/conversion/transformers.py
@@ -62,7 +62,7 @@ class Converter:
         return converted_models
 
     @staticmethod
-    def convert_from_transformers(model_name_or_path, device, task_type=None, processor=None):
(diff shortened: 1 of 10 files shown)


Timoeller (approved):
LGTM

tholor (approved):
Looking good. Can you maybe add one test case to verify that revisions are working (e.g. load  roberta-squad2 once in v1 and once in v2 and verify some of the weights)?

brandenchan:
I've fixed a few tests but some won't pass until we bump to 3.5.1 with #624

PR #649, opened 2020-12-04
WIP: Simplify processors - add Fasttokenizers

Timoeller:
Simplifying the processor by:
- moving functions into dataset_from_dicts
- unnesting functions
- cleaning up code

------------------

Some older commits are done by Bogdan and me, for making FARM work with transformers 3.5.1 and fasttokenizers.

For descriptions about progress see comments below.


Timoeller:
270996c introduces Multiprocessing after the Multithreading by Rust tokenizers. 

Although Multithreading should be finished by then, forking processes in python results in:
```
huggingface/tokenizers: The current process just got forked, after parallelism has already been used. Disabling parallelism to avoid deadlocks...
To disable this warning, you can either:
	- Avoid using `tokenizers` before the fork if possible
	- Explicitly set the environment variable TOKENIZERS_PARALLELISM=(true | false)
```

When setting the env var to false, python mp wont start.

My next experiment will be adding mp on a higher level (calling dataset_from_dicts) again.

Timoeller:
I did some performance benchmarking and **found the culprit**:
The function offset_to_token_idx() in samples.py takes up 93% of compute time.
Vectorizing the function reduced preprocessing from 199 secs to 21 secs!

---------------------

I will now work on separating the create_samples_qa and sample_to_features_qa functions into more meaningful functions:
1. split_documents_into_passages
2. featurize_text_and_labels

Timoeller:
Merged Brandens changes to the TextClassificationProcessor into this branch and also latest master.

Timoeller:
Separated NQ from SQuAD, so we can focus on SQuAD refactoring only.
NQ only works with slow tokenizers for now.

Timoeller:
QA processing tested by running question_answering_accuracy.py - using the Evaluator works as expected.

Working on inference mode now - this is still broken.

Timoeller:
Inference does not really work as of now, it seems the answer is moved by 1 token to the right.

Tested #624 where we updated to transformers 3.5.1 and added fasttokenizers. There the inferencer works.

Timoeller:
Found the bug! 
The bug was not only affecting inference but the whole QA processing pipeline:
We should not add special tokens when tokenizing documents and questions. These special tokens are added when qs and docs are combined.

Timoeller:
Refactored sample to features for QA.

QA processing tested by running question_answering_accuracy.py - Evaluator and Inferencer work as expected.

farm/modeling/prediction_head.py
```diff
@@ -690,41 +690,24 @@ def formatted_preds(self, logits, initial_mask, samples, return_class_probs=Fals
...
                 seq_res.append(
                     {
-                        "start": span["start"],
-                        "end": span["end"],
+                        "start": span[0],
+                        "end": span[1],
                         "context": f"{context}",
                         "label": f"{tag}",
-                        "probability": prob,
+                        "probability": np.float32(0.0),
```
Timoeller, line 710: @brandenchan 
is the prob completly gone or was this just a quick fix that you forgot to revert?

Timoeller:
I tested this branch with haystack in the following ways:

- running all haystack tests
- Tutorial 1 
- Tutorial 5, when using FarmReader.eval we need to adjust the return of processor.dataset_from_dicts
- Colab tutorial 5

Working on fixing the remaining tests.

Timoeller:
I cannot reproduce the failing s3 test and nothing has changed code wise.

I presume it is some CI problem that we can fix later.

Merging now.

farm/modeling/prediction_head.py
brandenchan: The old probability calculation was wrong. I have opened issue #658 to address this.

Issue #616, opened 2020-11-04
Remove mlflow logger for inference


tholor:
Don't initialize any mlflow run when we just initialize / use the inferencer. This is very annoying in production setups...

brandenchan:
Solved by #650

PR #650, opened 2020-12-04
MLFlowLogger can be disabled

brandenchan:
Related to #616.

To disable MLFlow logging, you simply call `MLFLowLogger.disable()`. ML Logging is now turned off in the Inferencer by default.

### DIFF
--- a/farm/utils.py
+++ b/farm/utils.py
@@ -95,7 +95,7 @@ class BaseMLLogger:
     This class can be extended to implement custom logging backends like MLFlow, Tensorboard, or Sacred.
     """
 
-    disable = False
+    disable_logging = False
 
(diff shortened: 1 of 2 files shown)


farm/utils.py
```diff
@@ -95,6 +95,8 @@ class BaseMLLogger:
     This class can be extended to implement custom logging backends like MLFlow, Tensorboard, or Sacred.
     """
 
+    disable = False
```
Timoeller, line 98: is that needed?

Timoeller (changes requested):
Made a code comment.

Apart from that, many tests are failing, then it OOMs with code 137. Can this be related to these changes?

Timoeller (approved):
LGTM

PR #651, opened 2020-12-04
Refactor Document Classification processing for Fast Tokenizers

brandenchan:
Refactor Document Classification Processor so that the fast Rust Tokenizers are first class citizens

### DIFF
--- a/farm/data_handler/processor.py
+++ b/farm/data_handler/processor.py
@@ -608,7 +608,8 @@ class TextClassificationProcessor(Processor):
 
         return dicts
 
-    def dataset_from_dicts(self, dicts, indices=None, return_baskets=False, inference=False, debug=False):
+    def dataset_from_dicts(self, dicts, indices=None, return_baskets=False, debug=False):
+        self.baskets = []
         # Tokenize in batches
         texts = [x["text"] for x in dicts]
         tokenized_batch = self.tokenizer.batch_encode_plus(
@@ -639,7 +640,8 @@ class TextClassificationProcessor(Processor):
                          "segment_ids": segment_ids}
 
             # Create labels
-            if not inference:
+            # i.e. not inference
+            if not return_baskets:
                 label_dict = self.generate_labels(dictionary)
                 feat_dict.update(label_dict)
 
--- a/farm/data_handler/processor.py
+++ b/farm/data_handler/processor.py
@@ -608,66 +608,98 @@ class TextClassificationProcessor(Processor):
 
         return dicts
 
-    def _dict_to_samples(self, dictionary: dict, **kwargs) -> [Sample]:
-        # this tokenization also stores offsets and a start_of_word mask
-        text = dictionary["text"]
-        tokenized = tokenize_with_metadata(text, self.tokenizer)
-        if len(tokenized["tokens"]) == 0:
-            logger.warning(f"The following text could not be tokenized, likely because it contains a character that the tokenizer does not recognize: {text}")
-            return []
-        # truncate tokens, offsets and start_of_word to max_seq_len that can be handled by the model
-        for seq_name in tokenized.keys():
-            tokenized[seq_name], _, _ = truncate_sequences(seq_a=tokenized[seq_name], seq_b=None, tokenizer=self.tokenizer,
-                                                max_seq_len=self.max_seq_len)
-        return [Sample(id=None, clear_text=dictionary, tokenized=tokenized)]
-
-    def _sample_to_features(self, sample) -> dict:
-        features = sample_to_features_text(
-            sample=sample,
-            tasks=self.tasks,
-            max_seq_len=self.max_seq_len,
-            tokenizer=self.tokenizer,
-        )
-        return features
-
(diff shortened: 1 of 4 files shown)


brandenchan:
Model trained with examples/doc_classification.py reaches 74.43% f1_macro

brandenchan:
Currently this branch is not passing the tests when run on a local GPU machine

farm/data_handler/processor.py
```diff
@@ -608,66 +608,98 @@ def file_to_dicts(self, file: str) -> [dict]:
...
-                      "offsets": offsets,
-                      "start_of_word": start_of_word}
+            # TODO Build tokenized dict for debug mode
+            tokenized = {}
 
-        if len(token_dict["tokens"]) == 0:
-            logger.warning(f"The following text could not be tokenized, likely because it contains a character that the tokenizer does not recognize: {text}")
-            return []
+            # Truncate results
+            token_ids = self.truncate(token_ids)
```
tholor, line 631: Why do you use a custom truncate method instead of supplying max_seq_len in batch_encode_plus? Just because of truncation stats?

farm/data_handler/processor.py
```diff
@@ -608,66 +608,98 @@ def file_to_dicts(self, file: str) -> [dict]:
...
+                         "segment_ids": segment_ids}
+
+            # Create labels
+            if not inference:
+                label_dict = self.generate_labels(dictionary)
+                feat_dict.update(label_dict)
+
+            self.add_basket(dictionary, tokenized, feat_dict)
 
+        if 0 in indices:
```
tholor, line 648: Not sure why we use this condition here and if it really triggers the logging as we expect it

farm/data_handler/processor.py
```diff
@@ -608,66 +608,98 @@ def file_to_dicts(self, file: str) -> [dict]:
...
+            if self._check_sample_features(basket):
+                for sample in basket.samples:
+                    features_flat.extend(sample.features)
+            else:
+                # remove the entire basket
+                basket_to_remove.append(basket)
+        dataset, tensor_names = convert_features_to_dataset(features=features_flat)
+        return dataset, tensor_names
+
+    def add_basket(self, dictionary, tokenized, feat_dict):
```
tholor, line 671: I would vote for moving this code directly into `dataset_from_dicts`. It's not that much of code and just reading "add_basket" in the code above is very cryptic.

farm/data_handler/processor.py
tholor: I think padding could also be handled by the new tokenizers directly

farm/data_handler/processor.py
```diff
@@ -608,66 +608,98 @@ def file_to_dicts(self, file: str) -> [dict]:
...
-
-    def _dict_to_samples_and_features(self, dictionary: dict, **kwargs) -> [Sample]:
-        """This method is used so that we need to tokenize only once when using a fast tokenizer."""
-        text = dictionary["text"]
-        inputs = self.tokenizer.encode_plus(
-            text=text,
-            max_length=self.max_seq_len,
-            truncation=True,
-            add_special_tokens=True,
+    def dataset_from_dicts(self, dicts, indices=None, return_baskets=False, inference=False, debug=False):
```
tholor, line 611: We should document the reasoning / behaviour of `inference` flag here.

tholor (commented):
Looks nice and way cleaner than before! Left a few comments ...

farm/data_handler/processor.py
brandenchan: Great point. Didn't realise new tokenizers could handle all that. The change is in the latest commit

farm/data_handler/processor.py
brandenchan: Currently it is there because we still call `dataset_from_chunk`. With the conditional, we only print a single sample. Without it, we print for every chunk. If we get rid of dataset chunking, we will probably be able to remove this conditional

farm/data_handler/processor.py
brandenchan: Done in latest commit

farm/data_handler/processor.py
brandenchan: After a short discussion with @Timoeller, we decided not to change the signature of `datset_from_dicts( )` yet since changing it for one task will require it to be changed for all. `return_baskets` is a good enough proxy for `inference` for now. We will change that naming at a later stage

farm/data_handler/processor.py
brandenchan: Hmmmm it seems as though this condition has different implications for each task. Need to investigate further

brandenchan:
Some tests are still failing.

e.g. 
The following comes from trying to call a Fast Tokenizer method while using a Slow Tokenizer for tests
`NotImplementedError: return_offset_mapping is not available when using Python tokenizers.To use this feature, change your tokenizer to one deriving from transformers.PreTrainedTokenizerFast.`

An error seems to be thrown when trying to log a random sample in lm finetuning

PR #655, opened 2020-12-14
Fix bug in save method of TextSimilarityProcessor

bogdankostic:
Passage tokenizer was assigned the name of the query tokenizer.


Timoeller (approved):
Looks ok.
We are refactoring the processors anyway, then we can deal with proper query_tokenizer and passage_tokenizer usage.

PR #656, opened 2020-12-14
Refactor NER for fast tokenizers

brandenchan:


brandenchan:
NER processing time on CONLL 03 reduced from 7.43s to 3.43s

Timoeller (approved):
Only touches NERs dataset from dicts. Looking much easier to understand.

I currently do not see an easy way to add checks for problematic samples. Lets add this potentially later, if bugs related to input data should surface or during our next input object refactoring.

PR #659, opened 2020-12-18
Refactoring Processor for LM Finetuning (FastTokenizers)

tholor:
Moving to fast tokenizers and new preprocessing stages ...

Breaking change: Slow tokenizers won't be supported any more for this type of task


Timoeller (approved):
Looks good.

Please also fix the lm finetuning tests (if they should be failing).

We need to make get start of word better in a coming release (branden also has a version as function of a single processor).

PR #662, opened 2020-12-23
Fix OOM in tests

Timoeller:
Tests exit with code 137 - OOM.

Sometimes OOMs in QA test, seen it once in test_ner_amp as well.

Trying to reduce memory usage inside test.

Also introducing rust multithreading parameter in inferencer.


tholor:
Merged and continued in #665

PR #663, opened 2020-12-24
Replace 'none' titles with empty string in DPR

tholor:
DPR preprocessing crashed in cases where `embed_title=True` but the incoming passages did not have a title.
Now, we replace those titles with an empty string and log a warning...


farm/data_handler/processor.py
```diff
@@ -3216,11 +3216,19 @@ def _dict_to_samples(self, dictionary: dict, **kwargs) -> [Sample]:
...
-                           zip(positive_ctx_titles, positive_ctx_texts)] + \
-                          [tuple((title, ctx)) for title, ctx in
-                           zip(hard_negative_ctx_titles, hard_negative_ctx_texts)]
+                # concatenate title with positive context passages + negative context passages
+                def _combine_title_context(titles, texts):
+                    res = []
+                    for title, ctx in zip(titles, texts):
+                        if title is None:
+                            title = ""
+                            logger.warning(
```
Timoeller, line 3225: This warning would be triggered for every datapoint if the title is missing.

Timoeller (approved):
I think this code makes FARM DPR more failsafe, so lets merge it.

We will need to refactor TextSimilarityProcessor and move the processing into dataset_from_dicts, as done with the other processors. Lets do this in a new PR

Timoeller:
Mhh dpr tests still failing, will need to investigate

PR #672, opened 2021-01-05
Added parameter names to convert_from_transformers call in question_answering_crossvalidation.py

julian-risch:
Parameter names were missing, a change in the order of parameters caused a bug. 
There were no other calls of this method that were missing parameter names.


Timoeller (approved):
LGTM.

I also checked the Inferencer.load call which has the new revision arg - seems fine for all occurrences as well.

Issue #673, opened 2021-01-05
datahandler.processor does not covert hundreds of samples to features


julian-risch:
**Describe the bug**
The question_answering_crossvalidation.py example discards hundreds of samples and does not convert them to features. Best guess is that there is a problem with the document striding and the tokenizer. Maybe it has been introduced with recent transformer updates.

**Error message**
`ERROR - farm.data_handler.processor -   Unable to convert 577 samples to features. Their ids are : 119-0-0, 127-1-4, ...` and hundreds of warnings like the following:
```
WARNING - farm.data_handler.processor -   Answer using start/end indices is ' 7.5 days' while gold label text is ' 7.5 days'.
Example will not be converted for training/evaluation.
```
Although the gold label text matches the text at the start/end indices, the sample is not converted.
**Expected behavior**
All samples with gold label text matching text at start/end indices should be converted to features.

**To Reproduce**
Steps to reproduce the behavior
Run question_answering_crossvalidation.py from examples of current master branch 1f046cadf9e5090d93a9760593b9788e1fe012c1
@Timoeller reproduced the behavior on a different machine. 

**System:**
 - OS: ubuntu
 - GPU/CPU: GPU 
 - FARM version: 0.6.0

Timoeller:
Fixed, closing now.

PR #679, opened 2021-01-07
Fix/missing truncation bug

julian-risch:
The tokenizer's call of the tokenize_batch_question_answering method caused warnings of untruncated input sequences. These warnings are now ignored because the encoded sequences are processed in strides afterward and truncation is therefore not needed.

Further, an input validation step for QA data had a bug where whitespaces were not correctly handled when comparing gold label texts with texts at the start/end indices. Leading and trailing whitespaces are now ignored

#673

### DIFF
--- a/farm/modeling/tokenization.py
+++ b/farm/modeling/tokenization.py
@@ -567,7 +567,7 @@ def tokenize_batch_question_answering(pre_baskets, tokenizer, indices):
     baskets = []
     # # Tokenize texts in batch mode
     texts = [d["context"] for d in pre_baskets]
-    tokenized_docs_batch = tokenizer.batch_encode_plus(texts, return_offsets_mapping=True, return_special_tokens_mask=True, add_special_tokens=False)
+    tokenized_docs_batch = tokenizer.batch_encode_plus(texts, return_offsets_mapping=True, return_special_tokens_mask=True, add_special_tokens=False, verbose=False)
 
     # Extract relevant data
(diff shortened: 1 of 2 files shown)


farm/modeling/tokenization.py
```diff
@@ -567,7 +567,7 @@ def tokenize_batch_question_answering(pre_baskets, tokenizer, indices):
     baskets = []
     # # Tokenize texts in batch mode
     texts = [d["context"] for d in pre_baskets]
-    tokenized_docs_batch = tokenizer.batch_encode_plus(texts, return_offsets_mapping=True, return_special_tokens_mask=True, add_special_tokens=False)
+    tokenized_docs_batch = tokenizer.batch_encode_plus(texts, return_offsets_mapping=True, return_special_tokens_mask=True, add_special_tokens=False, verbose=False)
```
Timoeller, line 570: Isnt there a parameter that controls padding explicitly? 
When this parameter is set to false it should not print the warning. When it still prints the warning this might be a bug in transformers (you can open an issue there once you have checked that for the current version of transformer + tokenizer)

Timoeller (commented):
Looks good.
Still there might be a logging bug on transformer side that we should investigate.

farm/modeling/tokenization.py
Timoeller: Seems like wanted logging behavior by transformers.

Timoeller (approved):
LG

PR #680, opened 2021-01-07
Added filter_range parameter that allows to filter answers with similar start/end indices

### LINKED ISSUE #667: Add filter for similar QA predictions
We want to be able to filter duplicate or near duplicate QA predictions.

julian-risch:
**Design choices**
Current question answering predictions contain (near-)duplicates. To ensure a variety of answer options coming from different text positions, this PR introduces a filtering step during the generation of predictions. To control the filtering, there is now an integer filter_range class variable for the class QuestionAnsweringHead. It is applied in the method get_top_candidates.
The default behavior is unchanged and corresponds to filter_range set to -1 (or smaller). Setting the parameter filter_range to 0 removes exact duplicates (same start or end index). Setting the parameter filter_range to any larger value consider answers with similar start or end index as duplicates, e.g., filter_range=5 considers the two answers with start_idx 4 and start_idx 9 as duplicates.

**Tests added**
test_duplicate_answer_filtering() tests whether there are no two generated answers with the same start or end index.
test_no_duplicate_answer_filtering() tests whether the default behavior is unchanged so that the answers contain duplicates.
test_range_duplicate_answer_filtering() tests whether filter_range = 5 leads to answers with a distance between start indices or end_indices of at least 6.

**Limitations**
If filter_range is to large (e.g., as large as the number of tokens in the given context) only one answer will be generated.
The similarity of answers is solely defined based on their start and end indices and does not consider similar answer texts with different indices as duplicates.

**Future enhancements**
Take into account textual similarity rather than only the indices. For example, compare exact words instead of start and end indices. For more advanced solutions, one could use locality-sensitive hash functions on the text of the generated answers and define a threshold of accepted hash distance.

Closes #667


examples/question_answering_filtering_similar_answers.py
```diff
@@ -0,0 +1,24 @@
+from farm.infer import QAInferencer
```
Timoeller, line 1: This might pollute our example section a bit much for such a small feature.

We can discuss if we want to introduce example folders based on tasks to organize the examples better.

farm/modeling/prediction_head.py
```diff
@@ -947,6 +948,10 @@ def __init__(self, layer_dims=[768,2],
                                   This is decoupled from n_best on document level, since predictions on passage level are very similar.
                                   It should have a low value
         :type n_best_per_sample: int
+        :param filter_range: The maximum distance that the start indices or end indices of two answers need to have to be handled as duplicates.
```
Timoeller, line 951: Maybe simplify naming and docstring?

Name the parameter: "duplicate_filtering" and the description "Answers are filtered based on their position. Both start and end position of the answers are considered. The higher the value answers that are more apart are filtered out."
Then give the exmaples.

Timoeller (changes requested):
Looking good already, lets discuss the proposed changes.

test/test_question_answering.py
```diff
@@ -140,9 +140,106 @@ def test_id(span_inference_result, no_answer_inference_result):
...
+        {
+            "questions": ["“In what country lies the Normandy?”"],
+            "text": """The Normans (Norman: Nourmands; French: Normands; Latin: Normanni) were the people who in the 10th and 11th centuries gave their name to Normandy, a region in France. They were descended from Norse (\"Norman\" comes from \"Norseman\") 
+                raiders and pirates from Denmark, Iceland and Norway who, under their leader Rollo, agreed to swear fealty to King Charles III of West Francia. Through generations of assimilation and mixing with the native Frankish and Roman-Gaulish populations, their descendants would gradually merge with the Carolingian-based cultures of West Francia. 
+                The distinct cultural and ethnic identity of the Normans emerged initially in the first half of the 10th century, and it continued to evolve over the succeeding centuries. Weird things happen in Normandy, France."""
+        }]
+
+    base_LM_model = "deepset/bert-base-cased-squad2"
+
+    inferencer = QAInferencer.load(base_LM_model, batch_size=2, gpu=False, task_type="question_answering",
```
Timoeller, line 154: Lets use fixtures to speed up the CI

julian-risch:
As discussed, I removed the example, used fixtures to speed up the CI, and renamed the parameter.

test/test_question_answering.py
```diff
@@ -140,9 +140,90 @@ def test_id(span_inference_result, no_answer_inference_result):
...
+    offset_answer_ends = []
+    for answer in result[0]["predictions"][0]["answers"]:
+        offset_answer_starts.append(answer["offset_answer_start"])
+        offset_answer_ends.append(answer["offset_answer_end"])
+
+    assert len(offset_answer_starts) == len(set(offset_answer_starts))
+    assert len(offset_answer_ends) == len(set(offset_answer_ends))
+
+
+def test_no_duplicate_answer_filtering(bert_base_squad2):
```
Timoeller, line 167: I wonder how this test can actually pass, since we already have exact duplicate filtering in place. 
It is used in the PredictionHead when we call formattedpreds to go from Pytorch predictions to actual String Answers.
See https://github.com/deepset-ai/FARM/blob/master/farm/modeling/prediction_head.py#L1386

Timoeller: I think here and also in the other tests it would make more sense to list the expected Answers. That way the testing is more explicit, easy to understand and does not open possibilities for unwanted side effects (e.g. tests passing although they should not pass)

Timoeller (commented):
Thanks for the improvements. I actually looked a bit deeper into the testing and found potential for improvement.

test/test_question_answering.py
julian-risch: Interesting. Okay, I will have a closer look and change the tests accordingly.

julian-risch:
The tests now check for the exact start and end indices so that they are more explicit.
The reason why the no_dupliacte_answer_filtering() test passed is because answers only need to have the same start indices OR the same end indices to count as duplicates here. The current method in the PredictionHead checks for the same start AND the same end indices to filter out duplicates.

Timoeller (approved):
LGTM

PR #694, opened 2021-01-21
Use Path instead of String in ONNXAdaptiveModel

skiran252:
load_dir should be a Path variable

### DIFF
--- a/farm/modeling/adaptive_model.py
+++ b/farm/modeling/adaptive_model.py
@@ -642,6 +642,7 @@ class ONNXAdaptiveModel(BaseAdaptiveModel):
 
     @classmethod
     def load(cls, load_dir, device, **kwargs):
+        load_dir = Path(load_dir)
         import onnxruntime
         sess_options = onnxruntime.SessionOptions()
         # Set graph optimization level to ORT_ENABLE_EXTENDED to enable bert optimization.


Timoeller (commented):
Thanks for the addition.

It currently only affects the ONNXAdaptiveModel. Did you check in our "normal" AdaptiveModel as well? If the changes and tests there become too complex, we can also move forward with just this PR.

Timoeller:
Again thanks for the addition, really appreciate effort coming from external developers. Some feedback for future OSS contributions though:
I changed the title of the PR - it is good practice to make the title more informative. 
Next time please also add slightly more description in the PR body, e.g. what caused the error or even link it to an open issue with a bug report.
Thanks

skiran252:
> Again thanks for the addition, really appreciate effort coming from external developers. Some feedback for future OSS contributions though:
> I changed the title of the PR - it is good practice to make the title more informative.
> Next time please also add slightly more description in the PR body, e.g. what caused the error or even link it to an open issue with a bug report.
> Thanks

Ya sure. thanks for the suggestion.

skiran252:
> Thanks for the addition.
> 
> It currently only affects the ONNXAdaptiveModel. Did you check in our "normal" AdaptiveModel as well? If the changes and tests there become too complex, we can also move forward with just this PR.

looks like this just affects ONNXAdaptiveModel and should be a simple change if done on that class.

Timoeller (approved):
OK, then lets merge.

PR #697, opened 2021-01-26
Re-enable test_dpr_modules also for windows.

ftesser:
Hello this PR is related to the closes isuue #637.

@Timoeller: I tested in my CI pipeline that with pytorch version 1.7.1, and  test_dpr_modules succeed on Windows, so probably as you pointed out from version 1.7.0 DDP is supported also in windows.
So, now that FARM makes use of pytorch 1.7.* it is desiderable to re-eanble this test for all operating system.

### DIFF
--- a/test/test_dpr.py
+++ b/test/test_dpr.py
@@ -1,4 +1,3 @@
-import os
 import pytest
 import torch
 import logging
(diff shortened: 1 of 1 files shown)


Timoeller (approved):
Nice, good suggestion!

I suggest to keep it to the windows user to install latest pytorch (in our reqs.txt there is torch>1.5,<1.8)

Issue #658, opened 2020-12-16
Assign correct probabilities to NER predictions


brandenchan:
In `TokenClassificationHead.formatted_preds( )`, we filter out spans for non-entity words, but we do not yet filter out probabilities of non-entity words. This leads to a misalignment of predictions and probabilities.

PR #700, opened 2021-02-01
Fix NER probabilities

brandenchan:
Previously, the NER probabilities being returned were often probabilities for the wrong entities. This PR addresses this issue (#658 )

### DIFF
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -696,11 +696,11 @@ class TokenClassificationHead(PredictionHead):
         for preds_seq, probs_seq, sample, spans_seq in zip(
             preds, probs, samples, spans
         ):
-            tags, spans_seq = convert_iob_to_simple_tags(preds_seq, spans_seq)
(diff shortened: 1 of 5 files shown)


brandenchan:
Prior to this PR, NER inferencing produced a single unnested list of Named Entities as predictions. After this PR, it is not a list of lists where each inner list corresponds to an input sample (i.e. if there are 3 input samples at inference time, the predictions will be a list populated by three lists).

It seems odd to me that this was the case. @Timoeller Do you think this change might break anything that is downstream of FARM?

Timoeller:
Mhh, I do not see where we use FARM NER in downstream applications and also "NER inferencing produced a single unnested list of Named Entities as predictions" seems not useful in the first place. So lets go with your newly implemented approach.

Timoeller (commented):
LGTM.

How about adding a test case that checks if labels and probs are in line now (I think it is good practice to always add test cases when we bugfix).

PR #702, opened 2021-02-02
Refactor Regression and inference processors

### LINKED ISSUE #676: Refactor Processor for regression, DPR, plain embeddings
The processor for these 3 tasks is still in old style.

We want all major functionality inside dataset_from_dicts

Timoeller:
There are quite some changed lines of code, but the overall changes are rather simple:
1. We are now using inside Regression and InferenceProcessor the TextClassificationProcessor.dataset_from_dicts method and only adjust the way labels are converted
2. A bugfix for conversion of python float labels to torch tensors. Previously they were silently converted from float to long, we now have to explicitly check for a correct conversion. This might affect other parts in our code as well. @tholor could you check if we have other float values that are converted to torch tensors?
3. Smaller change in dataset.py for the inferenceprocessing, we now take the whole text as context and not tokens.
4. I moved QAProcessor into NQProcessor, since it was not used elsewhere

partly closes #676

### DIFF
--- a/farm/data_handler/dataset.py
+++ b/farm/data_handler/dataset.py
@@ -19,9 +19,14 @@ def convert_features_to_dataset(features):
     all_tensors = []
     for t_name in tensor_names:
         try:
-            cur_tensor = torch.tensor(
-                [sample[t_name] for sample in features], dtype=torch.long
-            )
+            if t_name == 'regression_label_ids':
+                cur_tensor = torch.tensor(
+                    [sample[t_name] for sample in features], dtype=torch.float32
+                )
+            else:
+                cur_tensor = torch.tensor(
+                    [sample[t_name] for sample in features], dtype=torch.long
+                )
         except ValueError:
             cur_tensor = torch.tensor(
                 [sample[t_name] for sample in features], dtype=torch.float32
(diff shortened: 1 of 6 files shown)


Timoeller:
I now added checks weather data is silently converted to torch.long and tested most of our example scripts. I also added a couple of issues because some example scripts are currently not working.

@tholor please review

farm/data_handler/dataset.py
```diff
@@ -18,14 +23,30 @@ def convert_features_to_dataset(features):
...
+                    else:
+                        base = 1
+                else:
+                    base = features[0][t_name].ravel()[0]
+                if not np.issubdtype(type(base), np.integer):
+                    logger.warning(f"Problem during conversion to torch tensors:\n"
+                                   f"A non-integer value for '{t_name}' with a value of: "
+                                   f"'{base}' will be converted to a torch tensor of dtype long.")
+            except:
+                logger.warning("conversion checks did not work")
```
tholor, line 46: Let's make this warning here more descriptive. If users see this it should be clear where this comes from and what this means, e.g "Could not determine type for feature 'XXX'. Converting now to a tensor of default type long."

tholor (approved):
Looking good! I would only suggest changing one warning message

Issue #705, opened 2021-02-02
Model confidence 2.0


Timoeller:
As started in #690 we are developing a way to have better QA model confidence scores.

In this issue we especially focus on developing a method that is easily applicable to out of domain data. 

If it doesnt work in OOD data, create concept of what to do next.

Timoeller:
The linked PR 803 was unrelated to Model Confidence, so reopening issue again.

Issue #690, opened 2021-01-19
QA model confidence


Timoeller:
The QA answer score that is currently implemented can be misleading if not downright wrong. Lets create a "score" or confidence that is more informative.

The detailed plan how to approach QA model confidence follows soon.

Related issue on haystack side: https://github.com/deepset-ai/haystack/issues/739

julian-risch:
**First Plan**
We would like a QA model to state something like “with confidence of 75% the predicted answer is correct”.
Here comes a first simple plan based on a technique called temperature scaling. Further improvements could follow later on. The approach works not only for extractive QA but also for classification tasks. An adaptation of the approach might also work for Open Domain QA and RAG.

Current scores given for each generated answer combine the logits of the start and end index into one full answer score but do not really correspond to the model’s confidence. I suggest to scale (“calibrate”) these logits so that the scaled score corresponds to the accuracy (or other metrics) with which start and end indices are predicted correctly on a validation dataset. Thus, the model learns to output confidence scores that are calibrated on the observed confidence on a validation dataset.

The classification tasks (not QA), the idea is also described here: https://arxiv.org/pdf/1706.04599.pdf 
There is an implementation available here: https://github.com/gpleiss/temperature_scaling 

It comes down do adjusting the scores based on the following formula:
`softmax = e^(z/T) / sum_i e^(z_i/T)`
`T` is the temperature for scaling, learned based on the validation dataset
`z` is the logit for start index (or end index) as predicted by the model

For extractive QA, we would end up with two confidence scores, one for the start index and one for the end index. We need to try out different ways of combining them in one single score. 

@lalitpagaria  What do you think about the idea? Would you be interested in collaborating with us on this issue? We could schedule a call to discuss model confidence and the suggested plan. Integration with RAG seems to be more complex. Maybe right after the beam search step, when probabilities for each hypothesis are calculated, is a good point to use temperature scaling and output confidence scores.

lalitpagaria:
Sound good. I need to ramp up these can concept, I will read them and then we will schedule a call.

lalitpagaria:
Hi @julian-risch 
My apologies for late reply. I was held up by some other tasks. Unfortunately I will not get time, as my most of focus now a days on [Obsei](https://github.com/lalitpagaria/obsei). I would love to get your feedback about it.

I have gone through links and topics you shared I am able to understand them (at least Math part).
I see that you already started working on it, I will have better look into your PR to learn more about this topic.

[...]

PR #710, opened 2021-02-04
Model confidence calibration

julian-risch:
The score that is attached to `QACandidates` was not in the range [0,1] and was not aligned with the accuracy of the predictions.
This PR adds a confidence variable to `QACandidates` that lies in range [0,1] and is aligned with the accuracy of the predicted start position. For the alignment (calibration), we use temperature scaling on the dev set. To this end, `eval()` in the `Evaluator` class has a parameter `calibrate_conf_scores`. This parameter is false by default.

Example: After calibration, if the average confidence of 100 `QACandidates` is 70%, then the exact matches of start positions will also be on average 70%.

 #690 #705 

**Limitations:** 
* The calibration does not take into account end_position of the QA prediction. Only the start_position is used.
* The calibration does not work for language models that do not have the CLS token at position 0. For these models, we would need to map start_position==-1 to the index of CLS token in calibrate_conf().

**Future enhancements:** The calibration uses the exact match (EM) metric as a training target. On datasets where EM is low, we might want to use other metrics, e.g., F1 to be aligned with confidence scores.


Timoeller (commented):
Getting there, I left some comments along the way

examples/question_answering_model_confidence.py
```diff
@@ -0,0 +1,111 @@
...
+    logging.basicConfig(
+        format="%(asctime)s - %(levelname)s - %(name)s -   %(message)s",
+        datefmt="%m/%d/%Y %H:%M:%S",
+        level=logging.INFO,
+    )
+
+    ##########################
+    ########## Settings
+    ##########################
+    set_all_seeds(seed=42)
```
lalitpagaria, line 22: Do we need randomized seed here or user supplied?

Timoeller: I think Julian just took standard code from other examples and I believe the seed is rather irrelevant here.

farm/eval.py
```diff
@@ -37,7 +37,7 @@ def __init__(
         self.device = device
         self.report = report
 
-    def eval(self, model, return_preds_and_labels=False):
+    def eval(self, model, return_preds_and_labels=False, update_temp=False):
```
lalitpagaria, line 40: just suggestion to use `update_confidence` here as `temp` sometime used for temporary :)

julian-risch: Yes, agreed. I am thinking about a different naming. Maybe "calibrate_conf_scores"

farm/eval.py
```diff
@@ -90,6 +92,10 @@ def eval(self, model, return_preds_and_labels=False):
                 # TODO check why .fit() should be called on predictions, rather than on labels
                 preds_all[head_num] = mlb.fit_transform(preds_all[head_num])
                 label_all[head_num] = mlb.transform(label_all[head_num])
+            if head.model_type == "span_classification" and update_temp:
+                logger.info(f"temperature before update: {head.temperature}")
+                head.update_temperature(logits_all[head_num], label_all[head_num])
```
lalitpagaria, line 97: Same here as https://github.com/deepset-ai/FARM/pull/710/files#r570196858

farm/evaluation/metrics.py
```diff
@@ -156,7 +158,22 @@ def squad_EM(preds, labels):
...
-    return n_correct/n_docs
+    return n_correct/n_docs if n_docs else 0
+
+def squad_EM_start(preds, labels):
+    """
+    Count how often the predicted start index exactly matches the start index given by one of the labels
+    """
+    n_docs = len(preds)
+    n_correct = 0
+    for doc_idx in range(n_docs):
```
lalitpagaria, line 169: just suggestion, we could use `zip` as well -
```
for curr_labels, pred in zip(labels , preds):
```

farm/evaluation/metrics.py
```diff
@@ -189,11 +206,41 @@ def squad_f1_single(pred, label, pred_idx=0):
...
+
+def confidence(preds):
+    conf = 0
+    for pred in preds:
+        conf += pred[0][0].score
+    return conf/len(preds) if len(preds) else 0
+
+
+def metrics_per_bin(preds, labels):
+    pred_bins = [[] for _ in range(10)]
```
lalitpagaria, line 218: We can have variable called `num_of_bins=10` and use that for readability purpose.

farm/evaluation/metrics.py
```diff
@@ -189,11 +206,41 @@ def squad_f1_single(pred, label, pred_idx=0):
...
+    label_bins = [[] for _ in range(10)]
+    count_per_bin = [0]*10
+    for (pred, label) in zip(preds, labels):
+        current_score = pred[0][0].score
+        if current_score == 1.0:
```
lalitpagaria, line 223: Do we need to use `math.isclose` or `numpy` function with `eps =  10**-5`

farm/evaluation/metrics.py
```diff
@@ -189,11 +206,41 @@ def squad_f1_single(pred, label, pred_idx=0):
...
+            current_score = 0.9999
+        pred_bins[int(current_score*10)].append(pred)
+        label_bins[int(current_score*10)].append(label)
+        count_per_bin[int(current_score*10)] += 1
+
+    em_per_bin = [0]*10
+    confidence_per_bin = [0]*10
+    for i in range(1, 10):
```
lalitpagaria, line 231: Not sure are we intended to skip `index=0`?

farm/modeling/prediction_head.py
```diff
@@ -970,6 +970,7 @@ def __init__(self, layer_dims=[768,2],
         self.n_best_per_sample = n_best_per_sample
         self.duplicate_filtering = duplicate_filtering
         self.generate_config()
+        self.temperature = nn.Parameter(torch.ones(1) * 1)
```
lalitpagaria, line 973: Should we name it as `confidence`?

farm/modeling/prediction_head.py
```diff
@@ -1049,6 +1056,69 @@ def logits_to_loss(self, logits, labels, **kwargs):
...
+    def temperature_scale(self, logits):
+        """
+        Perform temperature scaling on logits
+        """
+        # Expand temperature to match the size of logits
+        if logits.dim() == 2:
+            temperature = self.temperature.unsqueeze(1).expand(logits.size(0), logits.size(1))
+        elif logits.dim() == 3:
+            temperature = self.temperature.unsqueeze(1).expand(logits.size(0), logits.size(1), logits.size(2))
+        return logits / temperature.to(device='cuda')
```
lalitpagaria, line 1068: not sure if we are intended to hardcode it for only GPU.

examples/question_answering_model_confidence.py
lalitpagaria: oh okay thanks for clarification. I don't have much idea thats why asked.

farm/eval.py
lalitpagaria: Yeah but yes instead of `conf` we can use full name `calibrate_confidence_scores`

farm/evaluation/metrics.py
julian-risch: You are right. It should be `range(0, 10)`

julian-risch: or better range(num_bins)

farm/evaluation/metrics.py
julian-risch: `current_score` should never be larger than 1 but I can change the if statement to `if current_score == 1.0:` to be sure.

farm/evaluation/metrics.py
julian-risch: Good idea. I will do that.

farm/evaluation/metrics.py
julian-risch: Good suggestion. It's more consistent to use `zip` here as well and I don't see any reason for not using it. 👍

farm/modeling/prediction_head.py
```diff
@@ -1348,18 +1391,25 @@ def reduce_preds(self, preds):
...
 
         # Iterate over the top predictions for each sample
         for sample_idx, sample_preds in enumerate(preds):
             best_pred = sample_preds[0]
             best_pred_score = best_pred.score
-            no_answer_score = self.get_no_answer_score(sample_preds) + self.no_ans_boost
+            best_pred_confidence = best_pred.confidence
+            no_answer_score, no_answer_confidence = self.get_no_answer_score_and_confidence(sample_preds)
+            no_answer_score += self.no_ans_boost
+            no_answer_confidence += self.no_ans_boost
```
Timoeller, line 1406: That does not seem right, since the boost can be anything from -inf to +inf

farm/modeling/prediction_head.py
```diff
@@ -1373,14 +1423,16 @@ def reduce_preds(self, preds):
...
+                                                        n_passages_in_doc=n_samples,
+                                                        confidence=qa_candidate.confidence)
                                             )
 
         # TODO add switch for more variation in answers, e.g. if varied_ans then never return overlapping answers
         pos_answer_dedup = self.deduplicate(pos_answers_flat)
 
         # This is how much no_ans_boost needs to change to turn a no_answer to a positive answer (or vice versa)
         no_ans_gap = -min([nas - pbs for nas, pbs in zip(no_answer_scores, passage_best_score)])
+        no_ans_gap_confidence = -min([nas - pbs for nas, pbs in zip(no_answer_confidences, passage_best_confidence)])
```
Timoeller, line 1435: I thought we would leave confidence as is and not adjust it in the complicated ways we do for the "normal" score

Timoeller (commented):
Looking good. I made some minor comments.

farm/modeling/prediction_head.py
julian-risch: Since I noticed that I did all my experiments and tests with this complicated method, I prefer to us this one. It works fine and has the advantage that the confidence is consistent with the score also for no_answers.

Timoeller (approved):
LGTM

Ierezell:
Hi, I also faced this problem and @PiffPaffM told me to look here. I wanted to know how I could solve that. 
I'm sorry to comment something that is barely related but it's my only entry point.... 

The problem I got is that the model can be really confident about the start position but the answer doesn't make sense
Like (not a real exemple but just to illustrate) : 

Q: "How high is the effeil Tower ?"  
Doc : "Some text ......   I have a small dog who really like to run outside .... Some more text" 
Answer : "small dog" `start_conf` : 0.89 / `end_conf` : 0.76 

So the score of QA would be high and I would think the answer could be found here. 

I tried to do cosine distance between the question and the span (or span +/- x words) to scale the confidence.
Also a keyword base approach and score based on ROUGE like metric. 
I also tried a topic modeling with LDA to check if answer and question were in the same topic (on haystack documents)
Even taking the best span of X documents and doing some sumarization etc... (on haystack retrieved documents)

Lot of custom metrics / number to improve the confidence and to get a "I didn't retrieve anything good" instead of a "You asked 5 documents here they are (irrevelant ones)" (for haystack but it's related here)

Again sorry if it was not the good place to ask. 

Thanks a lot for the nice work you're doing ! 

Have a great day

julian-risch:
Hi @Ierezell did you have a look at our following example? https://github.com/deepset-ai/FARM/blob/master/examples/question_answering_confidence.py Maybe it is helpful for you. This example shows how to tune confidence scores on a dev set. In particular, [line 92](https://github.com/deepset-ai/FARM/blob/816b4e3e65c142f8a31a63833058b75fe0419ed4/examples/question_answering_confidence.py#L92) might be interesting for you. The technique is called temperature scaling and is based on this [paper](https://arxiv.org/abs/1706.04599).
We will publish a blogpost early next week with a short explanation of these confidence scores. Stay tuned!

PR #713, opened 2021-02-10
Refactor Text pair handling

### LINKED ISSUE #688: SentencePair Regression
**Question**
Hi. Thank you for this wonderful repository. I am still in the early stages of using FARM.
I want to perform a sentencepair regression task with FARM. Basically I have two continues values for each sentence pair. I was going to use two regression heads for that. 

However, I noticed that There is no seperate processor for this. (TextPairRegressionProcessor), So I assumed that I have to use TextPairClassificationProcessor with task_type regression. 

Then I was examining the TextPairClassificationProcessor and noticed that it always checks for label_list. 

image

1. If my approach is right to use TextPairClassificationProcessor, what should I provide for label_list in the regression case?
2. If it is not will you be able to tell me the correct process to do text pair regression task?

Thank you in Advance and Thank you for this wonderful repo again

### LINKED ISSUE #701: Examples scripts using the TextPairClassificationProcessor are not working
**Describe the bug**
When trying to test our TextPairClassificationProcessor I realized a lot of our example scripts are currently not working.

## Example script text_pair_classification.py
- I think we have to upload the dataset "asnq_binary" for it to work. @brandenchan 
- inheriting from TextclassificationProcessor might be failing: in the dataset_from_dicts method that we now call directly from the TextclassificationProcessor there we just convert "text" and not "text_b" as well @Timoeller 

## Example script passage_ranking.py 
It throws an Exception("No task initialized....") @brandenchan

Timoeller:
PR fixes Text Pair classification by combining text and text_b in the proper way.

PR also adds text pair regression and accompanying tests.

fixes #701 
fixes #688

### DIFF
--- a/farm/data_handler/processor.py
+++ b/farm/data_handler/processor.py
@@ -10,30 +10,26 @@ from pathlib import Path
 from random import randint
 
 import numpy as np
-import torch
 from numpy.random import random as random_float
 from sklearn.preprocessing import StandardScaler
-from transformers import AutoConfig
 from tokenizers import Encoding
+from tokenizers.pre_tokenizers import WhitespaceSplit
+from transformers import AutoConfig
 
 from farm.data_handler.dataset import convert_features_to_dataset
-from farm.data_handler.input_features import get_roberta_seq_2_start, get_camembert_seq_2_start
 from farm.data_handler.input_features import sample_to_features_text
 from farm.data_handler.nq_utils import (
     sample_to_features_qa_Natural_Questions,
     create_samples_qa_Natural_Question,
     convert_qa_input_dict,
 )
-
 from farm.data_handler.samples import (
     Sample,
     SampleBasket,
     get_passage_offsets,
     offset_to_token_idx_vecorized
 )
-
 from farm.data_handler.utils import (
-    pad,
     expand_labels,
     read_tsv,
     read_tsv_sentence_pair,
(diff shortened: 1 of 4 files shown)


Timoeller:
@overlookcurl please have a look at this PR. There is also a new test for how to do TextpairRegression. I have tested it and it should work, at least the MSE went done rather splendidly.

We will merge this soon, if you do not want to wait you could test it out on this branch already :D

farm/data_handler/processor.py
```diff
@@ -721,6 +715,36 @@ def _create_dataset(self):
...
+            proxies=self.proxies,
+        )
+
+
+        # during tokenization inside TextClassificationProcessor only the "text" field is tokenized
+        # To convert text pairs we combine text and text_b into a tuple, then the tokenizer takes care of
+        # adding separator tokens and correct segment IDs
+        for d in dicts:
+            d["text"] = (d["text"], d["text_b"])
+        return dicts
```
tholor, line 745: Elegant trick! Shall we maybe add a test here to make sure TextPairClassificationProcessor.dataset_from_dicts() always contains correct segment IDs? Just in case huggingface's tokenizer signature changes or we go forward with our own Tokenizer class...might be a tricky bug to identify

tholor (approved):
Nice one! Just left one suggestion for a further test case...

farm/data_handler/processor.py
Timoeller: good idea, added.

PR #715, opened 2021-02-10
full dpr evalution added

tstadel:
`dpr_evaluation.py` script contains average rank evaluation over the full dpr dataset. In order to make this work some minor changes had to be made to `TextSimilarityHead` and `BaseBiAdaptiveModel`.

### DIFF
--- a/examples/dpr_evaluation.py
+++ b/examples/dpr_evaluation.py
@@ -1,0 +1,139 @@
+import logging
+import torch
+import gc
+import numpy as np
+from tqdm import tqdm
+
+from farm.data_handler.processor import TextSimilarityProcessor
+from farm.data_handler.utils import read_dpr_json
+from farm.modeling.biadaptive_model import BiAdaptiveModel
+from farm.modeling.tokenization import Tokenizer
+from farm.utils import set_all_seeds, initialize_device_settings
+from farm.infer import Inferencer
+
+def get_unique_queries(dpr_data):
+    return [{'query': query} for query in set(map(lambda q: q['query'], dpr_data))]
+
+def get_unique_passages(dpr_data):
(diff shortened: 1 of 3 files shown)


Timoeller:
Hey @tstadel thanks for the contribution. A better DPR evaluation is really needed to steer the training process.

We would take your code and make some adjustments so we can use the evaluation in our Evaluator object. See #716 
We will work on it in the coming sprint and will leave this PR open for now. Hope that is fine? Thanks again for this important feature.

tstadel:
Sure, you're welcome!

examples/dpr_evaluation.py
```diff
@@ -0,0 +1,140 @@
...
+    docs_tensor = torch.transpose(passage_tensors, 0, 1).cuda()
+    n_batches = n_data // batch_size + 1 if n_data % batch_size > 0 else 0
+    for idx in tqdm(range(n_batches)):
+        queries_t = []
+        target_idxs = []
+        start_idx = idx * batch_size
+        end_idx = min([idx * batch_size + batch_size, n_data])
+        actual_batch_length = end_idx - start_idx
+        for d in dpr_data[start_idx:end_idx]:
+            query_idx = query_to_idx[d['query']]
```
thenewera-ru, line 32: I suppose `query_to_idx` supposed to be passed as argument where you define it on line `134` ?

examples/dpr_evaluation.py
```diff
@@ -0,0 +1,140 @@
...
+            query_tensor = query_tensors[query_idx]
+            queries_t.append(query_tensor)
+            target_id = list(filter(lambda x: x['label'] == 'positive', d))[0]['external_id']
+            target_idx = passage_to_idx[target_id]
```
thenewera-ru, line 36: The same for `passage_to_idx`: did you mean to pass it as an argument as well otherwise it won't work?

PR #720, opened 2021-02-22
Adding an example showing how to calibrate and use model confidence scores

julian-risch:
This PR adds an example that shows how confidence scores can be calibrated and used with question answering models.
For example, predictions can be filtered out based on a confidence threshold.

#690 #705

### DIFF
--- a/examples/question_answering_confidence.py
+++ b/examples/question_answering_confidence.py
@@ -72,7 +72,10 @@ def question_answering_confidence():
     # The multiple predictions are used for evaluating top n recall.
     model.prediction_heads[0].n_best = accuracy_at
 
-    # 5. The calibration of model confidence scores sets one parameter, which is called temperature and can be accessed through the prediction_head
+    # 5. The calibration of model confidence scores sets one parameter, which is called temperature and can be accessed through the prediction_head.
+    # This temperature is applied to each logit in the forward pass, where each logit is divided by the temperature.
+    # A softmax function is applied to the logits afterward to get confidence scores in the range [0,1].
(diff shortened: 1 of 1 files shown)


examples/question_answering_confidence.py
```diff
@@ -0,0 +1,123 @@
...
+    # The multiple predictions are used for evaluating top n recall.
+    model.prediction_heads[0].n_best = accuracy_at
+
+    # 5. The calibration of model confidence scores sets one parameter, which is called temperature and can be accessed through the prediction_head
+    logger.info(f"Parameter used for temperature scaling of model confidence scores: {model.prediction_heads[0].temperature_for_confidence}")
+
+    # 6a. We can either manually set the temperature (default value is 1.0)...
+    model.prediction_heads[0].temperature_for_confidence = torch.nn.Parameter((torch.ones(1) * 1.0).to(device=device))
+
+    # 6b. ...or we can run the evaluator on the dev set and use it to calibrate confidence scores with a technique called temperature scaling.
```
Timoeller, line 81: Can we add some doc how the updated temperature is used/inserted? Is it changing the param in place? How to insert this value (what is the value of it anyways? Is it printed) into a model loaded from huggingface?

Timoeller (approved):
LG, one comment for improving documentation clarity.
If this is resolved you can merge directly.

examples/question_answering_confidence.py
julian-risch: Okay, I add some comments. Regarding models from huggingface: In this example, the model is loaded via `AdaptiveModel.convert_from_transformers`, which works fine.

PR #721, opened 2021-02-22
Added comments on how temperature parameter is automatically applied and stored

julian-risch:
This PR adds comments to document:
1. how the temperature parameter is automatically applied and updated when running the calibration and 
2. how it is automatically stored/loaded as an attribute of the prediction_head when storing/loading a model. 

#690 #705


Timoeller (commented):
LG

PR #725, opened 2021-02-24
fixed typo

dg4271:


Timoeller (approved):
Thanks for the fix. 
Although DPRReaders are currently not supported (and are non trivial to implement into farm) we should still send a proper error message.

Issue #730, opened 2021-03-03
Handling large file in DPR model 


voidful:
In DPR data loading https://github.com/deepset-ai/FARM/blob/5ab5b1620cb51ceb874d4b30c887e377ad1a6e9a/farm/data_handler/utils.py#L186

It read the entire JSON file at once and it would occur memory issue when reading a large file.

**Describe the solution you'd like**
Changing it to support both JSON and jsonl file determent by the input file type.
jsonl format is more suitable for handling large files. 
we can also provide a dpr_to_jsonl for conversion.

Timoeller:
Hey  @voidful seems like a good idea. 
Though just reading from JSONL would not solve the memory issue.
The whole file (weather it is json or jsonl) is loaded into python dicts and then converted into pytorch tensors (by multiple processes).

I see two options:
- For LM pretraining we have a StreamingDataSilo that does not convert the whole dataset at once. I shortly tested it for DPR processing, it does not work out of the box unfortunately. You could try fixing this.
- For training DPR only few/one negative passage is used. You could remove other unused passages from the file and drastically reduce its size. We would highly appreciate this downsampling in our `read_dpr_json` method.

Would any of those solve your issues?

voidful:
> Hey @voidful seems like a good idea.
> Though just reading from JSONL would not solve the memory issue.
> The whole file (weather it is json or jsonl) is loaded into python dicts and then converted into pytorch tensors (by multiple processes).
> 
> I see two options:
> 
> * For LM pretraining we have a StreamingDataSilo that does not convert the whole dataset at once. I shortly tested it for DPR processing, it does not work out of the box unfortunately. You could try fixing this.
> * For training DPR only few/one negative passage is used. You could remove other unused passages from the file and drastically reduce its size. We would highly appreciate this downsampling in our `read_dpr_json` method.
> 
> Would any of those solve your issues?

In my situation, it's because json load takes too much memory, iterative reading and appending data can solve my problem. (128GB memory, 30GB Data).
After that, all seems to be fine. However, I agree using the above strategy can further reduce memory usage. 
https://github.com/deepset-ai/FARM/blob/5ab5b1620cb51ceb874d4b30c887e377ad1a6e9a/farm/data_handler/utils.py#L217

Timoeller:
Ok, makes sense.
Happy to accept a PR with the functionality. 

Otherwise I would highly recommend modifying your 30GB DPR file or did you hand select those 30GB of data?
[...]

PR #733, opened 2021-03-04
perf(DPR-Data-Read): Improve data reading process

voidful:
Refer to issue #730 

Improvement:
- Support jsonl format for loading large file.
- Downsampling data to reduce memory usage.
- Add passage id when not exist.

### DIFF
--- a/farm/data_handler/utils.py
+++ b/farm/data_handler/utils.py
@@ -216,13 +216,16 @@ def read_dpr_json(file, max_samples=None, proxies=None, num_hard_negatives=1, nu
         logger.info(f" Couldn't find {file} locally. Trying to download ...")
         _download_extract_downstream_data(file, proxies=proxies)
 
-    if "jsonl" in file:
+    if file.suffix.lower() == ".jsonl":
         dicts = []
         with open(file, encoding='utf-8') as f:
             for line in f:
                 dicts.append(json.loads(line))
     else:
         dicts = json.load(open(file, encoding='utf-8'))
+
+    if max_samples:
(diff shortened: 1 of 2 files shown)


farm/data_handler/utils.py
```diff
@@ -214,9 +215,15 @@ def read_dpr_json(file, max_samples=None, proxies=None):
     if not (os.path.exists(file)):
         logger.info(f" Couldn't find {file} locally. Trying to download ...")
         _download_extract_downstream_data(file, proxies=proxies)
-    dicts = json.load(open(file))
-    if max_samples:
```
Timoeller, line 218: Lets keep max samples sampling here.

farm/data_handler/utils.py
```diff
@@ -214,9 +215,15 @@ def read_dpr_json(file, max_samples=None, proxies=None):
...
-        dicts = random.sample(dicts, min(max_samples, len(dicts)))
+
+    if "jsonl" in file:
```
Timoeller, line 219: There is a failing test:

```
if "jsonl" in file:
E       TypeError: argument of type 'PosixPath' is not iterable
```
Since file can also be of type PosixPath.
Lets check the file ending with `file.suffix.lower() == ".jsonl"`

Timoeller (commented):
Looks very good already. Thanks for the changes.
- lets keep max samples
- made a comment about a failing test

Although the downsampling now exists twice in our codebase I vote for keeping the downsampling in our TextSimilarityProcessor as wll, since we might use it for other datasets. Or do you have any strong preferences?

Timoeller (approved):
Very nice, thanks for the suggested changes.

PR #738, opened 2021-03-09
Fix reference before assignment in DataSilo

bogdankostic:
When using the DataSilo without train set (e.g. in evaluation), we get `UnboundLocalError: local variable 'ave_len' referenced before assignment` because `ave_len` and `clipped` are only defined for train sets. This PR sets these variables to a default value of `-1`if no train set is given.


Timoeller (approved):
LG

PR #744, opened 2021-03-24
WANDB logger support

thenewera-ru:
Any suggestions are welcome. 
@Timoeller, what do you say?

### DIFF
--- a/farm/utils.py
+++ b/farm/utils.py
@@ -227,6 +227,107 @@ class TensorBoardLogger(BaseMLLogger):
             TensorBoardLogger.summary_writer.add_text(tag=key, text_string=str(value))
 
 
+class WANDBLogger(BaseMLLogger):
+    """
+    WeightsANDBiases logger. See <a href="https://docs.wandb.ai">here</a> for more details.
+    """
+
+    experiment = None
+    save_dir = None
+    offset_step = 0
+    sync_step = True
+    prefix = ''
+    log_checkpoint = False
+
+    @classmethod
+    def init_experiment(cls,
+                        experiment_name,
+                        project_name,
+                        api: Optional[str] = None,
+                        notes=None,
+                        tags=None,
+                        entity=None,
+                        save_dir: Optional[str] = None,
+                        offline: Optional[bool] = False,
+                        _id: Optional[str] = None,
+                        log_checkpoint: Optional[bool] = False,
+                        sync_step: Optional[bool] = True,
+                        prefix: Optional[str] = '',
+                        notebook: Optional[str] = 'FARM',
+                        **kwargs):
+        if offline:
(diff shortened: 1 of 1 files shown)


bogdankostic:
Thanks for your contribution!

Could you maybe add an example on how to use the WANDBLogger? I just tried it on the `doc_classification.py`example by replacing `ml_logger = MLFlowLogger(tracking_uri="https://public-mlflow.deepset.ai/")`with `ml_logger = WANDBLogger(tracking_uri="")`but neither the parameters nor the metrics were tracked.

thenewera-ru:
Hey, @bogdankostic!

I will add the example on doc classification next week until then could you please try to replace
```
ml_logger = WANDBLogger(tracking_uri="")
```
with 
```
ml_logger = WANDBLogger.init_experiment(
   project_name='<YOUR PROJECT NAME YOU CREATE ON wandb.ai>',
   experiment_name='<EXPERIMENT NAME YOU CREATE on wandb.ai>',
   api='<YOUR API KEY you can see on wandb.ai in your settings>'
)
```
Hope it helps!
If anything is unclear feel free to ask.

Timoeller:
Hey any progress on the WANDB support? I think we were missing an example so we can run and test the contribution? Would be happy to go on here.

thenewera-ru:
I'm here, in the mean time I will provide the ex. notebook. Tested for my use-case training on russian `SQUAD`  dataset [here](https://arxiv.org/abs/1912.09723).

P.S. `data_handler/utils.py` there're lots of auxiliary methods for reading custom dataset including `read_dpr_json(...)` of course. In every method `encoding="utf-8"` when opening|closing file. When reading in language(s) other than english there might be encoding problem(s). 
I myself faced the error opening russian SQUAD data from  [here](https://arxiv.org/abs/1912.09723) with `utf-8`. Fixed replacing it with `utf-8-sig`.
So the question is shouldn't it be changed to `utf-8-sig` or perhaps it was set for some reason ? 
Since every file encoded in `utf-8` can be easily decoded with no problems in `utf-8-sig`. The reverse it not true, unfortunately.

Timoeller:
Hey thanks for the example scripts. We will look into it.


## About utf-8-sig
Disclaimer: I am not a string encoding expert, but... : )
Reading the docs https://docs.python.org/3/library/codecs.html#encodings-and-unicode it says utf-8-sig is just 3 added BOM tokens at the beginning of the sequence to increase the chance of automatically determining the character encoding used. I do not see how this could improve Cyrillic decoding, since the actual decoding is still plain uft-8. Could you give more insights into why it would be useful? In the linked docs it says "In UTF-8, the use of the BOM is discouraged and should generally be avoided." so I would also try to avoid it in FARM.

PR #746, opened 2021-03-26
WIP: DPR - enable loading of custom BERT models

Timoeller:
Saving works as well but there are problems when loading.

Also a proper test for saving and loading the processor and model is missing


julian-risch:
Tested on my side with gbert. Works like a charm.

julian-risch (approved):
Tested on my side with gbert. Works like a charm.

PR #753, opened 2021-04-30
Fix conversion of BiAdaptiveModel

bogdankostic:
When converting a DPR model to transformers and then trying to load this model again, the weights cannot be correctly mapped. This PR fixes this issue.


Timoeller (approved):
Looks good, I wonder how it was working before... o_O

Issue #760, opened 2021-05-07
Add farm.__version__


johann-petrak:
Instead of putting the version just into setup.py, it would be great to make the version in use available through `farm.__version__` as it is common practice with Python packages. 

There are many ways to achieve this, one simple way is to have a file `farm/_version.py` which sets the variable `__version__` to the version string and then import that variable into `farm/__init__.py`

The setup.py script could then load the version from `farm/_version.py` as well something similar to:

```
import os
import re
def versionfromfile(*filepath):
    infile = os.path.join(here, *filepath)
    with open(infile) as fp:
        version_match = re.search(
            r"^__version__\s*=\s*['\"]([^'\"]*)['\"]", fp.read(), re.M
        )
        if version_match:
            return version_match.group(1)
        raise RuntimeError("Unable to find version string in {}.".format(infile))

# ...

setup(..., version=versionfromfile("farm/_version.py"), ...)

```

Whenever the version number needs to get changed, the `_version.py` file is the only place to change. 

(Also the version number could be set to something indiciating non-release versions at all times except when an actual version that gets released on pypi is getting tagged). 

I find this approach quite useful and easy to implement in my own packages e.g. [gatenlp](https://github.com/GateNLP/python-gatenlp) and a number of popular projects are using similar approaches AFAIK.

Timoeller:
We definitely want a farm.__version__!

I liked you approach of having a single point for changing the version and it also seems to be [common practice](https://stackoverflow.com/a/7071358). Would you like to create a PR for this?

Timoeller:
Thanks for the valuable addition.
Fixed by #761 

btw when you write "fixes #761" in the PR message it fixes the issue [automatically](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

PR #761, opened 2021-05-08
Add farm.__version__ tag

johann-petrak:
Implements the changes proposed in #760 
Also makes setup.py executable on *NIX (undo if not wanted). 

The version added is  `0.7.2+snapshot` as 0.7.1 seems to be the last released version, so 0.7.2 would be next, but it is handy to indicate a non-released version during development.

### DIFF
--- a/farm/__init__.py
+++ b/farm/__init__.py
@@ -1,6 +1,7 @@
 import logging
 
 import torch.multiprocessing as mp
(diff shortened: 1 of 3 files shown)


Timoeller (approved):
Splendid.

Thanks for the great addition.

PR #765, opened 2021-05-14
Workaround transformers overwriting model_type when saving dpr models

julian-risch:
For DPR models, transformers overwrites the model_type parameter specified in a model's config when saving the model:
https://github.com/huggingface/transformers/blob/64e78564a519cda2b4408803e2781c604e1e3bdd/src/transformers/configuration_utils.py#L626 
For DPR models with a non-BERT tokenizer, e.g., CamembertTokenizer as in haystack issue [#1046](https://github.com/deepset-ai/haystack/issues/1046) this prevents loading the correct tokenizer when the model is loaded again. For example, the model_type _camembert_ is specified in the config when loading the model from the model hub, but the model_type _dpr_ is saved.
To overcome this problem, we set the model_type of transformers.DPRConfig on-the-fly to the model_type as specified in the model's config file.

This PR corresponds to haystack PR [#1060](https://github.com/deepset-ai/haystack/pull/1060)


farm/modeling/language_model.py
```diff
@@ -290,7 +294,22 @@ def save(self, save_dir):
...
         )  # Only save the model it-self
-        torch.save(model_to_save.state_dict(), save_name)
+
+        state_dict = model_to_save.state_dict()
+        keys = state_dict.keys()
+
+        # If we save a question_encoder or a ctx_encoder of a dpr model that is not based on a standard BERT model,
+        # e.g., a Camembert model, we need to adjust the names of the model weights.
+        # This adjustment removes a prefix name that otherwise would prevent the weights from being loaded again.
+        if self.model.config.model_type !="dpr" and (model_to_save.base_model_prefix.startswith("question_") or model_to_save.base_model_prefix.startswith("ctx_")):
```
Timoeller, line 304: Could we not move this part to the DPRQuestion- and ContextEncoder class by adding a save method there and calling the superclass' save method with altered state dicts? Should work, as mentioned here: https://discuss.pytorch.org/t/how-to-change-the-name-of-the-weights-to-a-new-name-when-saving-the-model/32166/2

Timoeller (approved):
Nice one.

I like the parametrization and documentation of the tests (since they can be quite confusing).
You even fixed the typo in the docs about cty_encoder : )

I added a comment that would make the solution even nicer (I believe). I would be fine with merging already now.

Timoeller:
:heart: ;)

PR #774, opened 2021-05-20
Fixes #773

### LINKED ISSUE #773: String comparison with "is"
See https://github.com/deepset-ai/FARM/blob/0357634c1997fe39b6f6e78827d35e43538caab1/farm/modeling/adaptive_model.py#L583

Should replace "is" with "=="

I do not think this is likely to cause a problem, but it depends on the way how Python string interning works.

johann-petrak:
Easy fix.


Timoeller (approved):
I dont think it matters much here, but lets merge

PR #782, opened 2021-05-27
Added warning and docstring about squad multiple answers management

ftesser:
@Timoeller this follow https://github.com/deepset-ai/FARM/issues/768

This PR contains 
 - some docstrings that describe the management of multiple answers;
 - some warnings  that notify the user when there are multiple answers or more answers than max_answers.

PR #783, opened 2021-05-28
Added assertion for questions with token length > self.max_query_length.

ftesser:
In Squad, I noticed that no log message or errors are thrown when questions are too long.
As my opinion this can cause wrong predictions, so better assert the right condition.

### DIFF
--- a/farm/data_handler/processor.py
+++ b/farm/data_handler/processor.py
@@ -1984,6 +1984,13 @@ class SquadProcessor(Processor):
             if basket.raw["document_text"] == "":
                 logger.warning("Ignoring sample with empty context")
                 continue
+            # error: question has a token length > than self.max_query_length
+            if len(basket.raw["question_tokens"]) > self.max_query_length:
+                str_err = "Question <{}> has a token length of {} greater than self.max_query_length({}).".format(basket.raw["question_text"],
+                                                                                                                 len(basket.raw["question_tokens"]),
+                                                                                                                 self.max_query_length)
+                logger.error(str_err)
+                assert False, str_err
             ########## end checking
 
 


ftesser:
Adding the assertion causes the fail of two tests (test_question_answering.test_training, test_question_answering.test_save_load) this because the question in the test is too long:

https://github.com/deepset-ai/FARM/pull/783/checks?check_run_id=2694268160#step:6:104

```AssertionError: Question <In what country is Normandy located?> has a token length of 7 greater than self.max_query_length(6).```

**Comment by ftesser** [comment]
Commit 5451d55 sets `max_query_length=7` to successfully pass the tests.

File: farm/data_handler/processor.py [hunk 1]
```diff
@@ -1984,6 +1984,13 @@ def _split_docs_into_passages(self, baskets):
             if basket.raw["document_text"] == "":
                 logger.warning("Ignoring sample with empty context")
                 continue
+            # error: question has a token length > than self.max_query_length
+            if len(basket.raw["question_tokens"]) > self.max_query_length:
+                str_err = "Question <{}> has a token length of {} greater than self.max_query_length({}).".format(basket.raw["question_text"],
+                                                                                                                 len(basket.raw["question_tokens"]),
+                                                                                                                 self.max_query_length)
+                logger.error(str_err)
+                assert False, str_err
```
Timoeller, line 1993: I like the idea of creating a message, maybe an info message? 
Lets not throw an exception that will make the program crash.

ftesser: I was thinking that a too long question was not correctly interpreted and therefore this should be reported as an error (even with assertion).

At this point I ask you: what happens when the question is longer than max_query_length?

julian-risch: If a question is longer than max_query_length, only the first max_query_length tokens are processed and the rest is cut off. In your example, the question `In what country is Normandy located?` would be shortened to `In what country is Normandy located` or to `In what country is Normandy`. In both cases, chances are high that we still get the correct answer. That's why I'd say an info message is enough and no assertion needed. 
Another reason for not using an assertion: if you switch from one language model to another, the tokenizer might change and so might the number of tokens in a given question. As a user, I would hardly understand why the code runs with one language model but a different model throws an exception for the same question.

ftesser: Thanks @julian-risch for the explanation, but what if max_query_length is 2 and the question is 20 long? 

I understand that the assertion is too heavy, but what about a warning message (instead of info message)?

julian-risch: You're right that we would not get the correct answer if max_query_length is 2 tokens and the question is 20 tokens long. Maybe a warning log message is a good fit here . However, I fear that this message might pollute the logging for large datasets with a relatively small amount of too long questions. For example, if a dataset of one million questions contains only 1% too long questions, there will still be 10000 warning messages. Could we warn the user only once maybe?

Timoeller: I think if a user is that advanced by setting these values its also in the users responsibility to use them sensible.
I strongly agree with Julian about reducing the number of log messages for irrelevant messages.

Also we need to have the assertionerror removed. Consider the case of a dataset with 1 million questions, you do not want it to break for one single malformed question.

ftesser: > Could we warn the user only once maybe?
this can be done using an additional member of the class (e.g. self.not_already_warned_about_query_lenght), if you like a solution like this I can made a commit for that
 
> I think if a user is that advanced by setting these values its also in the users responsibility to use them sensible.

I understand your point @Timoeller, but I think that a user (also an expert user) can forget/make mistake about all the needed setting parameters, for this reason I think it is better to implement these kind of check somewhere. I think FARM is the right place where this check can be done because FARM knows about the question_tokens lengths. 
What do you think? 

Meanwhile I pushed a version without assertion and with warning.

@julian-risch @Timoeller 

If you like to go for a solution with a member variable that remember if a log of this kind was already notify, i can add another commit; otherwise if you do not like warn the user at all or you have other idea let me know.

ftesser: I was thinking ... another solution might be to use some sort of tip's summary like what you did here: 



```
[Farmer's Tip] 83.7% of your samples got cut down to 512 tokens. Consider increasing max_seq_len. This will lead to higher memory consumption but is likely to improve your model performance
```

https://github.com/deepset-ai/FARM/blob/c9235f287ed92e09a46d7dc3572bf462f2ec95f7/farm/data_handler/data_silo.py#L471


This can be a solution to decrease the number of logging also in the case of  #782

PR #784, opened 2021-05-28
Added value of doc_stride, max_seq_len, max_query_length in error message

ftesser:
I am experimenting with different max_seq_len and max_query_length and I think it is nice to have information about these value in case of negative assertion.


Timoeller (approved):
Nice, thanks for the addition

PR #789, opened 2021-06-01
fixed typo and added import in code snippet

### LINKED ISSUE #788: Very first example in Readme.rst won't run
The very first example a new user may want to run is the code snippet in the README, but it does not mention the pathlib and there is a bracket missing.

Can we add the line:

    from pathlib import Path

and close the Path() call with a bracket?

I think it's quite an obstacle for curious new users like me when the first code snippet fails due to unclear assumptions and typos.

julian-risch:
Added import statement to code snippet and closed Path() call with a bracket.

Thanks to @nothingTVatYT for spotting these errors.

fixes #788


Timoeller (approved):
LG

PR #799, opened 2021-06-09
Shuffle data between epochs in DataLoader

Timoeller:
Currently we do not shuffle the data between epochs. 

I dont think this is meaningful in any case, so lets set the default mode to shuffling.


Timoeller:
We actually pass the DataLoader a RandomSampler or DistributedSampler. So we should keep the default value for shuffling=False.

PR #803, opened 2021-06-16
Semantic answer eval

### LINKED ISSUE #705: Model confidence 2.0

Timoeller:
As mentioned in #705 it uses output of the inferencer to compare prediction and labels with a sentencetransformer paraphrasing model.

fixes #705

### DIFF
--- a/farm/evaluation/semantic_answer_similarity_evaluation.py
+++ b/farm/evaluation/semantic_answer_similarity_evaluation.py
@@ -1,0 +1,102 @@
+from sentence_transformers import SentenceTransformer
+from sklearn.metrics.pairwise import cosine_similarity
+import numpy as np
+from farm.evaluation.metrics import top_n_accuracy
+
+
+def semantic_answer_similarity(result, sts_model_path_or_string="paraphrase-xlm-r-multilingual-v1", debug=False):
+    """
+    Computes BERT based similarity of prediction to gold labels.
+    Returns per QA pair a) the similarity of the most likely prediction to all available gold labels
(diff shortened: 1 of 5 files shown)


farm/evaluation/semantic_answer_similarity_evaluation.py
```diff
@@ -0,0 +1,109 @@
...
+        for j,p in enumerate(result[i].prediction):
+            p.meta = {"semantic_answer_score":np.max(sims[j, :])}
+
+        # Add debug information
+        if debug:
+            # Exclude no answer labels from debug info
+            if len(result[i].ground_truth_answer) == 0:
+                pass
+            else:
+                current_info = _add_debug_info(result=result, i=i, sims=sims)
```
julian-risch, line 68: `current_info` needs to be defined outside `if debug:` so that we can always return `current_info`

Timoeller: Would be a good catch but we return debug_info which is defined outside.

Timoeller: Ah the error comes from within the helper function. Will fix it.

julian-risch (approved):
LGTM! As discussed, an option for future improvements is to use the semantic evaluation only for samples that are not exact matches. For exact matches, we already know the similarity score -> 1.

Issue #800, opened 2021-06-10
BertStyleLMProcessor not working with german bert model


DieseKartoffel:
**Describe the bug**
I try to finetune a german bert model with my own text corpus. When I attempt to load the data in, the process will crash at `assert 103 not in tokens #mask token`. I was able to fix the issue by using "bert-base-cased" as the base model. However, I need to finetune "bert-base-german-cased". I tried changing changing other parameters but found that the model name is the cause for the problem, as I am able to reproduce the same error when using the lm_finetune.py from the farm examples and only changing the model name there.

**Error message**
```
>> python lm_finetuning.py 
/home/vagrant/anaconda3/envs/nlp/lib/python3.9/site-packages/torch/cuda/__init__.py:52: UserWarning: CUDA initialization: Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx (Triggered internally at  /pytorch/c10/cuda/CUDAFunctions.cpp:100.)
  return torch._C._cuda_getDeviceCount() > 0
06/09/2021 17:20:28 - INFO - farm.modeling.prediction_head -   Better speed can be achieved with apex installed from https://www.github.com/nvidia/apex .

 __          __  _                            _        
 \ \        / / | |                          | |       
  \ \  /\  / /__| | ___ ___  _ __ ___   ___  | |_ ___  
   \ \/  \/ / _ \ |/ __/ _ \| '_ ` _ \ / _ \ | __/ _ \ 
    \  /\  /  __/ | (_| (_) | | | | | |  __/ | || (_) |
     \/  \/ \___|_|\___\___/|_| |_| |_|\___|  \__\___/ 
  ______      _____  __  __  
 |  ____/\   |  __ \|  \/  |              _.-^-._    .--.
 | |__ /  \  | |__) | \  / |           .-'   _   '-. |__|
 |  __/ /\ \ |  _  /| |\/| |          /     |_|     \|  |
 | | / ____ \| | \ \| |  | |         /               \  |
 |_|/_/    \_\_|  \_\_|  |_|        /|     _____     |\ |
                                     |    |==|==|    |  |
|---||---|---|---|---|---|---|---|---|    |--|--|    |  |
|---||---|---|---|---|---|---|---|---|    |==|==|    |  |
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
[...]

PR #804, opened 2021-06-17
Dynamic token IDs for _mask_random_words function in BertStyleLMProcessor

DieseKartoffel:
Tokens were previously hardcoded so only the default config of bert-base was compatible. If the vocab.txt of a model had PAD, SEP, CLS tokens in a different position, the processor would ignore the wrong tokens for whole-word-masking. If MASK was at the wrong position, the processor would crash.

Tokens IDs are now grabbed from the vocabulary directly and IDs are not hardcoded anymore. This should make any model compatible that has SEP, CLS, PAD and MASK anywhere in their vocab.txt.

Related Issue: https://github.com/deepset-ai/FARM/issues/800

### DIFF
--- a/farm/data_handler/processor.py
+++ b/farm/data_handler/processor.py
@@ -1706,10 +1706,20 @@ class BertStyleLMProcessor(Processor):
         :type max_predictions_per_seq: int
         :return: (list of int, list of int), masked tokens and related labels for LM prediction
         """
+
         # 1. Combine tokens to one group (e.g. all subtokens of a word)
+
+
+        # tokens can have different ids depending on the model
+        pad_token_id = self.tokenizer.vocab["[PAD]"]
+        cls_token_id = self.tokenizer.vocab["[SEP]"]
+        sep_token_id = self.tokenizer.vocab["[CLS]"]
+        mask_token_id = self.tokenizer.vocab["[MASK]"]
+
+        
         cand_indices = []
         for (i, token) in enumerate(tokens):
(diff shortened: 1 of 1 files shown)


julian-risch (changes requested):
Looks good to me! Clean and solid. 👍  One thing: There is still one line of code that mentions index 103. That should be changed to `mask_token_id`, shouldn't it?
https://github.com/deepset-ai/FARM/blob/6f9fad140ff5b57bc081d311f2e2c9f370311a7b/farm/data_handler/processor.py#L1751
Have you already tested the changes? If so, how? Maybe with this test case but the language model changed to bert-base-german-cased? https://github.com/deepset-ai/FARM/blob/d200c8c85ad6ba23307f6971de36da19cd8b3631/test/test_lm_finetuning.py#L18

DieseKartoffel:
oh yes thanks i overlooked that!
for now i only started the training process for testing and saw that the processor was returning valid data and not crash. wasnt able to finish a training and check results yet. any other ideas for testing this that could make sense?

julian-risch:
That's good! I did some local testing with`FARM/test/test_lm_finetuning.py` and changed the language model to bert-base-german-cased there. Works. I found out that the token "I" has index 103 in bert-base-german-cased vocabulary. So I also tested with this token by adding it to `FARM/test/samples/lm_finetuning/train-sample.txt`. It works with your changes but it does not without them. Great! 👍 Ready to merge. Let us know how the finetuned model performs!

julian-risch (approved):

PR #805, opened 2021-06-20
Update processor.py

guan27:
PTAL
fix log_problematic():
problematic_sample_ids is list of int rather than list of str, so I changed their variable type.


Timoeller (approved):
Well spotted, thanks!

PR #814, opened 2021-06-25
Torch Version 1.9 Compatibility check

### LINKED ISSUE #810: Allow torch 1.9 version in requirements.txt
Hi, is there reason for torch 1.9 to be excluded from requirements.txt dependencies? These days I am working on LFQA demos on Colab which comes with torch 1.9 pre-installed. Every time I install haystack pip picks up torch<1.9,>1.5 dependency from farm and it installs 1.8.1 torch, while 1.9 is already preinstalled. 

I've seen that you guys do torch compatibility testing for every new torch release. How likely are new torch releases to break farm? Would it be possible to leave upper bound on torch dependency open?

Timoeller:
fixes #810


julian-risch:
Tests passed (including text pair classification) and benchmarks passed (with a different seed). @Timoeller  I'd say this PR is ready to be merged.

julian-risch:
However, remember to change the fixed torch version to `torch>1.5,<1.10` before merging.

Issue #822, opened 2021-07-03
NER Inference always returns empty list


DieseKartoffel:
**Question**
Hello,

I trained an NER model on my own dataset (about 30k labeled sentences) and the evaluation looks very promising (Test set has about 93% accuracy on 18 different entity types). However, whenever I run inference, i always get an empty predictions list. I tried many examples from my train and test set and never got any prediction.

 This is not the case when I use the model that was trained with your [ner example](https://github.com/deepset-ai/FARM/blob/master/examples/ner.py). Here inference works fine. I used the debugger to compare the processed data of your example with mine and it looks the same (key is "text" with the correct "ner_label" list). 

I was able to replicate the problem by creating a simplified dataset with 25 examples and a single entity type. I overfit the model on this dataset to 100% accuracy and try to run inference on an example from the training data. Here is how the data format looks like:

```
-DOCSTART-	O
Ich	O
bin	O
7	number
.	O

-DOCSTART-	O
Mein	O
Alter	O
ist	O
12	number
.	O
. . .
```

When I run:
```
    basic_texts = [
        {"text": "Mein Alter ist 12 ."},
    ]
    model = Inferencer.load(save_dir)
    result = model.inference_from_dicts(dicts=basic_texts)
   print(result)
```

I get:
```
   [{'task': 'ner', 'task_name': 'ner', 'predictions': [[]]}]
```

The Code I use for training is [here](https://github.com/DieseKartoffel/my-fileshare-repo/blob/master/farm_ner_train.py).
The only thing that is different from your working example is the dataset, the label_list, the delimiter and num_epochs. I also tried changing the delimiter and renaming the labels to fit to your example but it did not help.

[Here](https://paste.ubuntu.com/p/MFtyf4dC95/) is a full dump of the training process that shows correctly looking training samples, 100% accuracy on test set, followed by empty inference prediction on an example from the training set.

[Here](https://paste.ubuntu.com/p/fxQDYWBdqS/) is the simplified dataset I used to reproduce the problem.

I hope I am not overlooking something obvious here. I you have any idea for what else I could look into to fix this, it would be very much appreciated!

FabriceEll:
I came across this issue as well a couple months ago when using FARM in my Windows work environment, ended up not using NER because of this which kinda sucks

julian-risch:
[...]

PR #828, opened 2021-07-13
Checking for iob format when preparing ner results

DieseKartoffel:
Simple check if ner tags are actually in IOB format when presenting inference results. Before the results would just be empty if the data was not in IOB2. Now, if they are, carry on as before and `convert_iob_to_simple_tags`. If they are not, present prediction results for all of the input tokens. "O" labels are removed in any case.

This is related to issue https://github.com/deepset-ai/FARM/issues/822 and should help any user that is unaware of the IOB conventions by presenting corresponding warnings if the dataset is not IOB formatted.

Tbh, I am not sure if an exception in the NER processor (+ some additional documentation for NER) might not be a better idea than this approach. The user could be forced to always use IOB2 formatting and get some info about it. Also if a user has a dataset with IOB1 instead of IOB2 `convert_iob_to_simple_tags` might return wrong results. However, if FARM should be flexible as in accepting any NER dataset, than these warnings might be a better solution.

### DIFF
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -705,7 +705,28 @@ class TokenClassificationHead(PredictionHead):
         for preds_seq, probs_seq, sample, spans_seq in zip(
             preds, probs, samples, spans
         ):
-            tags, spans_seq, tag_probs = convert_iob_to_simple_tags(preds_seq, spans_seq, probs_seq)
+
+            # If user is working with correct IOB2 format, convert each B-I group to single label and remove any O
+            if any(tag.startswith(("I-", "B-")) for tag in preds_seq):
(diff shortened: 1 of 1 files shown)


julian-risch (approved):
@DieseKartoffel Looks good to me! 👍 
Regarding the TODO note: The `_convertIOB1_to_IOB2` function in data_handler/utils.py does not only do conversion but also checks for correct IOB1 format. 
https://github.com/deepset-ai/FARM/blob/81d1b31791a213be17709cf9d46f9c9981579e34/farm/data_handler/utils.py#L275
A check for correct IOB2 format would need to ensure that every I-tag is preceded by a B-tag of the same entity type (PER, LOC, ...)

Thank you very much for your contribution! 🙂

Issue #829, opened 2021-07-15
Remove unnecessary logging statement


johann-petrak:
There is a left-over logging statement in the `data_silo` code, line 431, remove it.
https://github.com/deepset-ai/FARM/blob/3e7ca55533e3614015e0cff51a4b09fb990db770/farm/data_handler/data_silo.py#L431

PR #830, opened 2021-07-15
Remove unnecessary logging statement, fixes issue #829

johann-petrak:


Timoeller:
I dont see the harm in that info but happy to remove it if you are bothered by it :D

PR #833, opened 2021-07-28
Use correct class when transforming DPR models

bogdankostic:
When trying to convert a DPR model to transformers, `DPRQuestionEncoder` is used for the context encoder model. This is due to the fact that transformer's `AutoModel` always creates a `DPRQuestionEncoder` when it is passed a `DPRConfig`. This PR ensures that we use `DPRContextEncoder` for the context encoder model.


julian-risch (commented):
LGTM! Ready to merge. Still a mystery though how it worked for GermanDPR in the past. 🤔

Timoeller:
Nice catch. What would be the consequence of using a DPRQuestionEncoder for the context?

PR #836, opened 2021-08-17
Add switch to QA pred head for ranking by confidence scores

### LINKED ISSUE #808: allow switching between confidence scores in reader
Let's have only one score in answers but support switching between the two options: `probability` and `score`
To this end, the prediction head needs to sort based on a switch here:
https://github.com/deepset-ai/FARM/blob/cf6f96ffcedeb7b1341278dcdd6437e99efaa75b/farm/modeling/prediction_head.py#L1467
Right now, answers contain both `probability` and `score`: 
```
{'answer': 'Eddard',
 'context': 's Nymeria after a legendary warrior queen. '
            'She travels with her father, Eddard, to '
            "King's Landing when he is made Hand of the "
            'King. Before she leaves,',
 'document_id': 'ba2a8e87ddd95e380bec55983ee7d55f',
 'meta': {'name': '43_Arya_Stark.txt'},
 'offset_end': 78,
 'offset_end_in_doc': 153,
 'offset_start': 72,
 'offset_start_in_doc': 147,
 'probability': 0.9899835586547852,
 'score': 15.927592277526855}

```

Related issue in Haystack https://github.com/deepset-ai/haystack/issues/1220

julian-risch:
`QACandidates` contain `score` and `confidence` fields but only `score` was used to rank `QACandidates` so far.
This PR makes ranking `QACandidates` by score the default but also allows to set `use_confidence_scores_for_ranking` in `QAPredictionHead`, which activates ranking 
`QACandidates` by `confidence`.

I think, it's better to keep both fields, `score` and `confidence` in `QACandidates` instead of dropping one of them entirely. Thereby, we are more flexible and allow backwards compatibility.

A new test case checks that results are ranked as expected with and without setting `use_confidence_scores_for_ranking`.

closes #808

### DIFF
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -964,6 +964,8 @@ class QuestionAnsweringHead(PredictionHead):
         :type duplicate_filtering: int
         :param temperature_for_confidence: The divisor that is used to scale logits to calibrate confidence scores
         :type temperature_for_confidence: float
+        :param use_confidence_scores_for_ranking: Whether to sort answers by confidence score (normalized between 0 and 1) or by standard score (unbounded)
+        :type use_confidence_scores_for_ranking: bool
         """
         super(QuestionAnsweringHead, self).__init__()
         if len(kwargs) > 0:
--- a/farm/modeling/prediction_head.py
+++ b/farm/modeling/prediction_head.py
@@ -941,6 +941,7 @@ class QuestionAnsweringHead(PredictionHead):
                  n_best_per_sample=None,
                  duplicate_filtering=-1,
                  temperature_for_confidence=1.0,
+                 use_confidence_scores_for_ranking=False,
                  **kwargs):
         """
         :param layer_dims: dimensions of Feed Forward block, e.g. [768,2], for adjusting to BERT embedding. Output should be always 2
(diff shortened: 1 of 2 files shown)


farm/modeling/prediction_head.py
```diff
@@ -963,6 +964,8 @@ def __init__(self, layer_dims=[768,2],
         :type duplicate_filtering: int
         :param temperature_for_confidence: The divisor that is used to scale logits to calibrate confidence scores
         :type temperature_for_confidence: float
+        :param use_confidence_scores_for_ranking: Whether to sort answers by confidence score (normalized between 0 and 1) or by standard score (unbounded)
```
tholor, line 967: Can we document here somewhere a bit clearer what these different scores exactly are and how they are calculated? If I remember correctly we have these three cases:
a) score = start + end logit (unbounded)
b) confidence (default) = logits scaled to 0-1 and incorporating no_answer 
c) confidence (calibrated) - same a b) but we multiply it with a learned scaling parameter
I am sure I will forget about it in a couple of weeks and would be helpful to have it documented for others. Probably it's best to do that in the general prediction head doc string (+ Haystack's FARMReader).

tholor (approved):
LGTM. 
Small thing: Not directly related to this PR but I think it's a great opportunity to add a bit more explanation around the different scores we have (see comment)

farm/modeling/prediction_head.py
julian-risch: Good point. I added an explanation of the three kinds of scores to the doc string of the QAPredictionHead.

PR #846, opened 2021-11-11
AdaptiveModel.convert_to_onnx does not save float16 model conversion to output_path

### LINKED ISSUE #845: AdaptiveModel.convert_to_onnx does not save float16 model conversion to output_path
I think there is a small bug with `AdaptiveModel.convert_to_onnx` when enabling float16 conversion:

https://github.com/deepset-ai/FARM/blob/7305a17979b0a80dbe2dbebe5815450883f20627/farm/modeling/adaptive_model.py#L614-L624

The resulting model file ends up in the current working directory instead of `output_path/model.onnx`.

Should be something like:

```
            optimized_model.save_model_to_file(str(output_path/"model.onnx"))
```

cjb06776:
Small fix to correct path to output converted ONNX model when enabling float16 conversion.

fixes #845

### DIFF
--- a/farm/modeling/adaptive_model.py
+++ b/farm/modeling/adaptive_model.py
@@ -621,7 +621,7 @@ class AdaptiveModel(nn.Module, BaseAdaptiveModel):
                 hidden_size=config.hidden_size
             )
             optimized_model.convert_model_float32_to_float16()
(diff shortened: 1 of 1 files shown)


Timoeller (approved):
LG, thanks for the fix

